PR #2, opened 2020-03-16
Lexer 

luke-bhan:
Ok I wrote the lexer in libslide folder, I am a little confused about why theres two separate projects. Also, I think the code could be really bad as its my first one, but I did test it thoroughly in my opinion. Let me know what you think! Sorry for the wait!


ayazhafiz:
There is one project for the library (libslide, where we will likely do most of our development) and one project for the app (likely a command-line application).

libslide/src/scanner.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+impl Scanner {
+    // instantiate a new scanner
+    pub fn new(input: &str) -> Scanner {
+        Scanner{
+            input: input.to_owned(),
+            output: Vec::new()
+        }
+    }
```
ayazhafiz:
```suggestion
    pub fn new<T: Into<String>>(input: T) -> Scanner {
        Scanner{
            input: input.into(),
            output: Vec::new()
        }
    }
```

This will allow passing a `&str`, `String`, or anything else

libslide/src/scanner.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    // matches token with symbol and creates it: private helper function
```
ayazhafiz:
```suggestion
    // matches token with symbol and creates it
```

libslide/src/scanner.rs
```diff
@@ -0,0 +1,233 @@
...
+            ')' => t = TokenType::CloseParen, 
+            '[' => t = TokenType::OpenBra, 
+            ']' => t = TokenType::CloseBra,
+             _  => t = TokenType::Empty
+        }
+        let ret = Token{token: t, ..Default::default()};
+        return ret;
+   }
+
+    // iterates through any digits to create a token of that value: private helper method
```
ayazhafiz:
```suggestion
    // iterates through any digits to create a token of that value
```

libslide/src/scanner.rs
```diff
@@ -0,0 +1,233 @@
...
+            '%' => t = TokenType::Mod,
+            '^' => t = TokenType::Exp,
+            '=' => t = TokenType::Equal,
+            '(' => t = TokenType::OpenParen, 
+            ')' => t = TokenType::CloseParen, 
+            '[' => t = TokenType::OpenBra, 
+            ']' => t = TokenType::CloseBra,
+             _  => t = TokenType::Empty
+        }
+        let ret = Token{token: t, ..Default::default()};
```
ayazhafiz: nice!

libslide/src/scanner.rs
```diff
@@ -0,0 +1,233 @@
...
+        return ret;
+   }
+
+    // iterates through any digits to create a token of that value: private helper method
+    fn iterate_digit(&mut self,mut i: usize,mut c: char) -> (Token, usize){
```
ayazhafiz: Maybe rename to `iterate_number` or `scan_number`?

libslide/src/scanner.rs
```diff
@@ -0,0 +1,233 @@
...
+
+
+            
+
+        
+
+
+
+        
+
```
ayazhafiz: please remove extra whitespace

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,144 @@
...
+    // empty token
+    Empty
+}
+
+// This will hold are token data
+#[derive(Copy, Clone, Debug)]
+pub struct Token {
+    pub token: TokenType,
+    pub integer: i64,
+    pub float: f64,
```
ayazhafiz: I think it might be better just to store the token value as a string. We can unwrap the actual token value later on. Otherwise if we add variables, then we will also have to include `string` as data member here.

ayazhafiz: Another option is to create multiple token types with one base algebraic type. What do you think?

libslide/src/scanner.rs
```diff
@@ -0,0 +1,233 @@
...
+        equal == a.len() && equal == b.len()
+    }
+
+    #[test]
+    fn test_empty_string_scan() {
+        let mut s = Scanner::new("");
+        s.scan();
+        let result = Vec::<Token>::new();
+        assert_eq!(true, compare_vec::<Token>(&result, &s.output));
+    }
```
ayazhafiz: These tests look good; I will probably replace them after this PR because I have been doing some research on alternative testing strategies that I think may be better suited for this kind of thing. But I will send a PR for that.

libslide/src/scanner/types.rs
luke-bhan: what does it mean to have a base algebraic type

libslide/src/scanner.rs
luke-bhan: ok

libslide/src/scanner/types.rs
luke-bhan: i was just gonna do it how you did it in rcalc

ayazhafiz: That’s what it means

luke-bhan:
ok take a look but should be fixed.

libslide/src/scanner.rs
```diff
@@ -0,0 +1,224 @@
...
+            '/' => t = TokenType::Div,
+            '%' => t = TokenType::Mod,
+            '^' => t = TokenType::Exp,
+            '=' => t = TokenType::Equal,
+            '(' => t = TokenType::OpenParen, 
+            ')' => t = TokenType::CloseParen, 
+            '[' => t = TokenType::OpenBra, 
+            ']' => t = TokenType::CloseBra,
+             _  => t = TokenType::Empty
+        }
```
ayazhafiz:
```suggestion
        let t = match c {
            '+' => TokenType::Plus,
            '-' => TokenType::Minus,
            '*' => TokenType::Mult,
            '/' => TokenType::Div,
            '%' => TokenType::Mod,
            '^' => TokenType::Exp,
            '=' => TokenType::Equal,
            '(' => TokenType::OpenParen, 
            ')' => TokenType::CloseParen, 
            '[' => TokenType::OpenBra, 
            ']' => TokenType::CloseBra,
             _  => TokenType::Empty
        };
```

libslide/src/scanner.rs
```diff
@@ -0,0 +1,224 @@
...
+        let mut dec_str = "0.".to_owned();
+        let ret: Token;
+        // iterate through integer part
+        while c.is_digit(10) && i < self.input.chars().count() {
+            int_str.push(c);
+            i += 1;
+            if i < self.input.chars().count() {
+                c = self.input.as_bytes()[i] as char;
+            }
+        }
```
ayazhafiz: ?? why not just iterate through `input.chars()` rather than assigning to c?

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,130 @@
+// A definition of a token - used in the scanner
+// Written by Luke Bhan, 2/19/2020
+//
+//
+#[derive(PartialEq, Copy, Clone, Debug)]
+pub enum TokenType{
+    // Stores a floating point number as dp
+    Num(f64),
```
ayazhafiz: Let's call this `Float`?

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,130 @@
...
+    // Identifies division - stored as a char
+    Div,
+
+    // Identifies modulo - stored as a char
+    Mod,
+
+    // Identifies exponent - stored as a char
+    Exp,
+
+    // Identifies an equal sign - stored as a char
```
ayazhafiz: Well actually you're not storing anything right

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,130 @@
...
+    OpenBra, 
+    
+    // close bracket ]
+    CloseBra,
+
+    // empty token
+    Empty
+}
+
+// This will hold are token data
```
ayazhafiz:
```suggestion
// This will hold token data
```

libslide/src/scanner.rs
```diff
@@ -0,0 +1,224 @@
...
+            c = self.input.as_bytes()[i] as char;
+            while c.is_digit(10) && i < self.input.chars().count(){
+                dec_str.push(c);
+                i += 1;
+                if i < self.input.chars().count() {
+                    c = self.input.as_bytes()[i] as char;
+                }
+            }
+            // turn integer and decmial strings into token
+            ret = Token{token: TokenType::Num(int_str.parse::<i64>().unwrap() as f64 + dec_str.parse::<f64>().unwrap())}
```
ayazhafiz: Why not just parse as f64?

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,130 @@
...
+    OpenParen,
+
+    // close parentheses )
+    CloseParen, 
+
+    // open bracket [
+    OpenBra, 
+    
+    // close bracket ]
+    CloseBra,
```
ayazhafiz:
```suggestion
    CloseBracket,
```

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,130 @@
...
+    Equal, 
+
+    // open parentheses (
+    OpenParen,
+
+    // close parentheses )
+    CloseParen, 
+
+    // open bracket [
+    OpenBra,
```
ayazhafiz:
```suggestion
    OpenBracket, 
```

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+impl Token{
+    pub fn is_empty(self) -> bool {
+        if TokenType::Empty == self.token{
+            return true;
+        }
+        else{
+            return false;
+        }
```
ayazhafiz: boolean zen my guy.
```suggestion
        self.token == TokenType::Empty
```

libslide/src/scanner.rs
```diff
@@ -0,0 +1,224 @@
...
+            '*' => t = TokenType::Mult,
+            '/' => t = TokenType::Div,
+            '%' => t = TokenType::Mod,
+            '^' => t = TokenType::Exp,
+            '=' => t = TokenType::Equal,
+            '(' => t = TokenType::OpenParen, 
+            ')' => t = TokenType::CloseParen, 
+            '[' => t = TokenType::OpenBra, 
+            ']' => t = TokenType::CloseBra,
+             _  => t = TokenType::Empty
```
ayazhafiz: Let's make this an invalid Token.

```rust
TokenType {
  ...
  Invalid(String)
}

File: libslide/src/scanner/types.rs [hunk 19]
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+impl Default for Token{
+    fn default() -> Token{
+        Token {token: TokenType::Empty}
+    }
+}
```
ayazhafiz: I don't think this makes semantic sense. What is a "default" token?

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,130 @@
...
+
+impl PartialEq for Token{
+    fn eq(&self, other: &Self) -> bool{
+        self.token == other.token
+    }
+}
```
ayazhafiz: Why not just derive this?

libslide/src/scanner/types.rs
ayazhafiz: Also if you change this to `Invalid` this function is probably not necessary.

libslide/src/scanner.rs
luke-bhan: why should invalid hold a string

luke-bhan:
also I dont think we even need a token struct then. Cant we just use the enum

luke-bhan:
thoughts @ayazhafiz

libslide/src/scanner.rs
```diff
@@ -0,0 +1,209 @@
...
+            '^' => TokenType::Exp,
+            '=' => TokenType::Equal,
+            '(' => TokenType::OpenParen, 
+            ')' => TokenType::CloseParen, 
+            '[' => TokenType::OpenBracket,
+            ']' => TokenType::CloseBracket,
+             _  => TokenType::Invalid(c.to_string())
+        };
+        let ret = Token{token: t};
+        return ret;
```
ayazhafiz:
```suggestion
        Token {token: t}
```

Let's also rename the `token` field to `type`.

libslide/src/scanner.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    pub fn scan(&mut self){
+        let mut i: usize = 0;
+        let mut c: char;
+        let mut t: Token;
+        let mut tuple: (Token, usize);
+        // iterate through string
+        while i < self.input.chars().count() {
+            // ignore whitespace
+            if !((self.input.as_bytes()[i] as char).is_whitespace()) {
```
ayazhafiz: Can you not do
```suggestion
            if !((self.input[i]).is_whitespace()) {
```
?

libslide/src/scanner.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+#[cfg(test)]
+mod tests{
+    use super::*;
+
+    fn compare_vec<T: PartialEq>(a: &Vec<T>, b: &Vec<T>) -> bool{
+        let equal = a.iter().zip(b.iter()).filter(|&(a,b)| a == b).count();
+        equal == a.len() && equal == b.len()
+    }
```
ayazhafiz: I think rust already supports direct comparison of two vectors..

libslide/src/scanner/types.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub token: TokenType
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_eq_1(){
+        let t = Token{token: TokenType::Invalid("s".to_string())};
```
ayazhafiz:
```suggestion
        let t = Token{token: TokenType::Invalid("s".into())};
```

libslide/src/scanner.rs
luke-bhan: no

luke-bhan: strings cannot be indexed as you'd expect in rust. They have to coverted to bytes. It is O(1) though so I think its ok

libslide/src/scanner.rs
luke-bhan: yes will fix.

libslide/src/scanner.rs
luke-bhan: type is a keyword. lets use token_type

libslide/src/scanner.rs
ayazhafiz: Oh that's weird. In this case, shouldn't we convert to as_bytes before we start the lexing?

luke-bhan: thatd probably be a cleaner way to do it

luke-bhan: so changing everything to as_bytes gets confusing
Basically as bytes returns a utf8 array. However, setting input to a utf8 array results in an unexpected size error. Therefore, we should write input as a reference to a utf8 array. However, this then gets into lifetime issues in structs. From there we would need to declare an explicit lifetime of our struct scan which seems very complex to just simplify as_bytes when as_bytes is O(1)

PR #5, opened 2020-03-23
Humanize scanner test strategy

### LINKED ISSUE #4: Lexer: Improve tests
Humanize tokens created in the scanner and normalize test cases.

ayazhafiz:
This commit improves tests for slide's scanner and tokens by converting
scanned tokens to a human-readable string form, and then comparing that
string form to a string expectation. With a macro that abstracts
creation of a scanner and translation of the program input to this
human-readable string, this test strategy should make it very easy to
see what tests there currently are, and add/modify more as necessary.

Closes #4.

### DIFF
--- a/libslide/src/scanner.rs
+++ b/libslide/src/scanner.rs
@@ -1,23 +1,23 @@
 mod types;
-use types::TokenType;
 use types::Token;
+use types::TokenType;
 
 pub struct Scanner {
-    input: Vec<char>, 
-    pub output: Vec<Token>
+    input: Vec<char>,
+    pub output: Vec<Token>,
 }
 
 impl Scanner {
     // instantiate a new scanner
     pub fn new<T: Into<String>>(input: T) -> Scanner {
-        Scanner{
+        Scanner {
             input: input.into().chars().collect(),
-            output: Vec::new()
+            output: Vec::new(),
         }
     }
 
-    // matches token with symbol and creates it: private helper function 
-   fn create_symbol_token(c: char) -> Token{
+    // matches token with symbol and creates it: private helper function
+    fn create_symbol_token(c: char) -> Token {
         let t = match c {
             '+' => TokenType::Plus,
             '-' => TokenType::Minus,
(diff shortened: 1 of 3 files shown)


ayazhafiz:
ptal @luke-bhan. there are some miscellaneous whitespace changes (my auto-formatter did this); if it is a burden, I can remove them and do the whitespace changes in another commit. Thanks!

libslide/src/scanner/types.rs
```diff
@@ -29,66 +32,88 @@ pub enum TokenType{
...
+                Equal => "=".into(),
+                OpenParen => "(".into(),
+                CloseParen => ")".into(),
+                OpenBracket => "[".into(),
+                CloseBracket => "]".into(),
+                Invalid(s) => format!("Invalid({})", s),
+            }
+        )
     }
+}
```
luke-bhan: why do we use into for plus, minus, mult versus from. Since char and string are naturally converted. I may not understand the difference between from and into fully, but I think we should be differentiating.

libslide/src/scanner/types.rs
```diff
@@ -29,66 +32,88 @@ pub enum TokenType{
...
+impl fmt::Display for Token {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use TokenType::*;
+        write!(
+            f,
+            "{}",
+            match &self.token_type {
+                Float(num) => num.to_string(),
+                Int(num) => num.to_string(),
+                Plus => "+".into(),
```
luke-bhan: why do we use into for plus, minus, mult versus from. Since char and string are naturally converted. I may not understand the difference between from and into fully, but I think we should be differentiating.

libslide/src/scanner.rs
```diff
@@ -1,23 +1,23 @@
...
-use types::TokenType;
 use types::Token;
+use types::TokenType;
 
 pub struct Scanner {
-    input: Vec<char>, 
-    pub output: Vec<Token>
+    input: Vec<char>,
+    pub output: Vec<Token>,
 }
```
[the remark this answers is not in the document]
ayazhafiz: I don't know. This is Rust style convention; presumably, it makes adding a new item to the end of the list easier because you don't have to add a comma to the one before

libslide/src/scanner/types.rs
ayazhafiz: This match arm returns a string, because of the Float and Int matches. So we have to convert a `char` or `&str` into a String. This is why we call `char.into()`. We could also do `String::from(char)`, but this is more typing.

luke-bhan: String::from(char) i think is rust convention though per https://doc.rust-lang.org/std/convert/trait.Into.html

luke-bhan: actually, I think it's fine.

ayazhafiz: Is it? The example uses into: https://doc.rust-lang.org/std/convert/trait.Into.html#examples

PR #6, opened 2020-04-12
Initial parser

ayazhafiz:


libslide/src/parser.rs
```diff
@@ -0,0 +1,63 @@
...
+    Int(i64),
+    BinOp(Box<BinOp>),
+    // I added un op even though I havent implemented + or -
+    UnaryOp(Box<UnaryOp>),
+}
+
+pub struct BinOp {
+    pub item: Token,
+    pub lhs: Expr,
+    pub rhs: Expr
```
ayazhafiz, line 14: It should be enough to make these `Box<Expr>`. Then you will be able to construct a `BinOp` with any expression.

libslide/src/parser.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+pub struct UnaryOp{
+    pub item: Token, 
+    pub rhs: Expr
```
ayazhafiz, line 19: same here
```suggestion
    pub rhs: Box<Expr>
```

libslide/src/parser.rs
```diff
@@ -0,0 +1,63 @@
+use crate::scanner::types::Token;
+
+pub enum Expr {
+    Float(f64),
+    Int(i64),
+    BinOp(Box<BinOp>),
+    // I added un op even though I havent implemented + or -
+    UnaryOp(Box<UnaryOp>),
```
ayazhafiz, line 8: You actually don't need `Box` here, you only need to hold it dynamically when the precise type of the expression is not known (i.e. in the left/right of a binary op, left comments below). So this can be
```suggestion
    BinOp(BinOp),
    // I added un op even though I havent implemented + or -
    UnaryOp(UnaryOp),
```

ayazhafiz:
@luke-bhan please see the comments above

luke-bhan:
hey! I added some code for +/- and *. Also parens and brackets should work. Do you wanna make a testing environment? I haven't tested the parser so I doubt it really works as of right now, but I know you'll prolly wanna define a print macro for our tree. @ayazhafiz

libslide/src/parser.rs
```diff
@@ -0,0 +1,135 @@
...
+
+pub struct UnaryOp{
+    pub item: Token, 
+    pub rhs: Box<Expr>,
+}
+
+pub struct Parser<'a>{
+    input: &'a Vec<Token>,
+    index: usize,
+    cur_token: Token,
```
ayazhafiz, line 26: If we have an index, why do we need a `cur_token`? If the worry is with borrowing, let's just own the input instead (`Vec<Token>`)

libslide/src/parser.rs
```diff
@@ -0,0 +1,135 @@
...
+        Parser{
+            input: input.into(),
+            index: 1,
+            cur_token: input[0].clone().into(),
+        }
+    }
+
+    fn get_token(&mut self) -> Token{
+        self.index = self.index+1;
+        return self.input[self.index-1].clone();
```
ayazhafiz, line 40: why not return a `&Token`?

libslide/src/parser.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+        return node;
+    }
+
+    fn num_term(&mut self) -> Box<Expr> {
+        // 5.0 is a placeholder. mem::discriminant only compares variant types and ignores data
+        // this is pretty fucking cool rust has it
+        // this value should never be returned
+        let mut node = Box::new(Expr::Int(1));
+        if std::mem::discriminant(&self.cur_token.token_type) == std::mem::discriminant(&TokenType::Float(5.0)) {
```
ayazhafiz, line 72: memory discrimination is pretty expensive. we can do the same thing in a more clear manner using pattern matching:

```rust
match self.cur_token.token_type {
  TokenType::Float(_) => ...
}
```

libslide/src/parser.rs
```diff
@@ -0,0 +1,135 @@
...
+        // 5.0 is a placeholder. mem::discriminant only compares variant types and ignores data
+        // this is pretty fucking cool rust has it
+        // this value should never be returned
+        let mut node = Box::new(Expr::Int(1));
+        if std::mem::discriminant(&self.cur_token.token_type) == std::mem::discriminant(&TokenType::Float(5.0)) { 
+            // look into a better way of doing this. only other method I can find is match
+            // by this I mean extract the data of variant float
+            if let TokenType::Float(f) = self.cur_token.token_type{
+                node = Box::new(Expr::Float(f));
+            }
```
ayazhafiz, line 77: if you do pattern matching in the way i show in the comment above you can get this from the matched expression.

libslide/src/parser.rs
```diff
@@ -0,0 +1,135 @@
...
+                self.cur_token = self.get_token();
+            }
+            return node;
+        }
+        else if self.cur_token.token_type == TokenType::OpenParen {
+            // eat left paren
+            self.cur_token = self.get_token();
+            node = self.expr();
+            // eat right paren
+            self.cur_token = self.get_token();
```
ayazhafiz, line 97: can you add TODO: check for errors here?

ayazhafiz:
@luke-bhan i've updated the PR with a test framework you can use

luke-bhan:
why did you get rid of expr, binaryop and unary op

luke-bhan:
@ayazhafiz

ayazhafiz:
@luke-bhan I didn't, I moved it to `types`

ayazhafiz:
Oh it didn't add it. Should be fixed now.

luke-bhan:
gotcha!

luke-bhan:
wait does the parser work? I haven't tested it yet but from what you wrote its passing all tests?

libslide/src/parser.rs
luke-bhan: the reason we have an index is to hold a variable to iterate through the vector and the cur_token. The reason we have a cur token is so that we don't need to keep passing the token from each recursive call since they all have access to self. This could defniitely be reworked i don't disagree

libslide/src/parser.rs
luke-bhan: so we can reassign it. If we don't then we have to borrow it which creates a mess. We can try this as well.

libslide/src/parser.rs
luke-bhan: ok I'm on it

ayazhafiz:
Yes. Run `cargo test`

luke-bhan:
take a look. I fixed the matching error and added the todo and added modulo. Do you wanna try and fix the cur_token with the indexing issue. also, we need to add exponent as well. I'll try to look at parens but my isd final is tomorrow so I might not be free till thursday. @ayazhafiz

ayazhafiz:
@luke-bhan sounds good, I might make some changes to this PR if that's good with you.

luke-bhan:
ofc

libslide/src/parser.rs
```diff
@@ -0,0 +1,138 @@
...
+            || self.cur_token.token_type == TokenType::Minus
+        {
+            let t = self.cur_token.clone();
+            self.cur_token = self.get_token();
+            node = Box::new(Expr::BinOp(BinOp {
+                item: t,
+                lhs: node,
+                rhs: self.mul_divide_mod_term(),
+            }));
+        }
```
ayazhafiz, line 42: to avoid a mut reference, we could have a match expression here return the value.

libslide/src/parser.rs
```diff
@@ -0,0 +1,138 @@
...
+                rhs: self.num_term(),
+            }));
+        }
+        return node;
+    }
+
+    fn num_term(&mut self) -> Box<Expr> {
+        // 5.0 is a placeholder. mem::discriminant only compares variant types and ignores data
+        // this is pretty fucking cool rust has it
+        // this value should never be returned
```
ayazhafiz, line 65: comment is unnecessary now right?

libslide/src/parser.rs
```diff
@@ -0,0 +1,138 @@
...
+        let node: Box<Expr>;
+        match self.cur_token.token_type {
+            TokenType::Float(f) => {
+                node = Box::new(Expr::Float(f));
+                if self.index < self.input.len() {
+                    self.cur_token = self.get_token();
+                }
```
ayazhafiz, line 72: This could be a helper

libslide/src/parser.rs
```diff
@@ -0,0 +1,138 @@
...
+                // eat left brac\ket{
+                self.cur_token = self.get_token();
+                node = self.expr();
+                self.cur_token = self.get_token();
+                return node;
+            }
+            _ => {
+            // this should never be reached
+            panic!("Invalid input");
+            },
```
ayazhafiz, line 101:
```suggestion
            _ => unreachable!()
```

ayazhafiz:
@luke-bhan let me know how this looks to you. If it's good, could you add support for parsing unary operators and tests for parens/brackets? I think we are good to merge the PR after that.

After this, I think we should add support for parsing symbols (like variables) and then we can get started on the fun part -- optimization :)

luke-bhan:
ok. looks better. we also need support for the exponent correct

ayazhafiz:
Do you want to implement that in this PR? Otherwise I can do it, or we can do it later.

luke-bhan:
doesn't matter. The only tricky part is the right associativity. I can do it

ayazhafiz:
cool sounds good. can you be sure to add the others i mentioned too?

luke-bhan:
ok, I am a little confused with the right associativity with unary and exponent. Do you want to exponent as an example and Ill do the unary ops after possibly?

luke-bhan:
wait nvm i got it

luke-bhan:
Ok, I added unary, expo and assignment. I didn't do a large amount of testing, but I think a few expressions. If you want to add more tests I get that. @ayazhafiz

libslide/src/parser.rs
```diff
@@ -0,0 +1,210 @@
...
+            TokenType::OpenBracket => {
+                // eat left bracket
+                self.advance();
+                let node = self.expr();
+                self.advance();
+                node
+            }
+            TokenType::Variable(_) => {
+                self.assignment()
+            }
```
ayazhafiz, line 141: But this won't work if we do something like `1 + a` now?

ayazhafiz: LGTM other than that, great job :+1: but we should fix this

ayazhafiz: Don't worry about robustness for now, we can just panic wherever, but we need to be able to support this so that we can start the optimization pipeline

ayazhafiz: If you want I can do it, just lmk

ayazhafiz: Other option is we can ignore this and just say we only parse one expression/statement for now, since that's what we do anyway. Otherwise we'll need to support multi-statement programs for variables in the RHS to be useful anyway. That is, with this parser

```
a = 1;
b = 1 + a;
```

is not supported anyway because we parse at most one expression.

ayazhafiz: Let me know what you think. I'm happy to just leave as is for now so we can start on the other parts of this program, and we can clean this up later.

ayazhafiz: @luke-bhan

luke-bhan: sounds good. Lets just leave it. Ready to merge I think right?

luke-bhan: I also completely forgot about that lol

luke-bhan: @ayazhafiz

PR #12, opened 2020-04-28
Evaluate

luke-bhan:
take a look at what i have. I have some issues with casting. We can implement functions for each enum variant to cast it from Expr to BinOp or we can try some other design ideas. Enum variants and the visitor pattern in rust are proving to mix rather poorly in my opinion. PTAL @ayazhafiz

### DIFF
--- a/libslide/src/partial_evaluator/types.rs
+++ b/libslide/src/partial_evaluator/types.rs
@@ -1,0 +1,68 @@
+use crate::parser::types::*;
+use core::fmt;
+
+pub enum PEResult {
+    Evaluated(f64),
+    Unevaluated(Box<Expr>),
+}
+
+impl PEResult {
+    pub fn from_expr(expr: Expr) -> Self {
+        Self::Unevaluated(Box::new(expr))
+    }
+
+    pub fn fold_binary(lhs: PEResult, rhs: PEResult, op: BinaryOperator) -> PEResult {
+        use PEResult::*;
+        let binary_fn = binary_operator_table(&op);
+        match (lhs, rhs) {
+            (Evaluated(x), Evaluated(y)) => Evaluated(binary_fn(x, y)),
+            (Unevaluated(lhs), Unevaluated(rhs)) => {
+                PEResult::from_expr(Expr::BinaryExpr(BinaryExpr { op, lhs, rhs }))
+            }
+
+            (Evaluated(x), y) => Self::fold_binary(PEResult::from_expr(Expr::Float(x)), y, op),
+            (x, Evaluated(y)) => Self::fold_binary(x, PEResult::from_expr(Expr::Float(y)), op),
+        }
+    }
+
+    pub fn fold_unary(rhs: PEResult, op: UnaryOperator) -> PEResult {
+        use PEResult::*;
+        let unary_fn = unary_operator_table(&op);
+        match rhs {
+            Evaluated(x) => Evaluated(unary_fn(x)),
+            Unevaluated(rhs) => PEResult::from_expr(Expr::UnaryExpr(UnaryExpr { op, rhs })),
+        }
+    }
+}
+
+fn binary_operator_table(op: &BinaryOperator) -> Box<dyn Fn(f64, f64) -> f64 + 'static> {
+    match op {
+        BinaryOperator::Plus => Box::new(|x, y| x + y),
+        BinaryOperator::Minus => Box::new(|x, y| x - y),
+        BinaryOperator::Mult => Box::new(|x, y| x * y),
+        BinaryOperator::Div => Box::new(|x, y| x / y),
+        BinaryOperator::Mod => Box::new(|x, y| x % y),
+        BinaryOperator::Exp => Box::new(|x, y| x.powf(y)),
+    }
+}
+
+fn unary_operator_table(op: &UnaryOperator) -> Box<dyn Fn(f64) -> f64 + 'static> {
+    match op {
+        UnaryOperator::SignPositive => Box::new(|x| x),
+        UnaryOperator::SignNegative => Box::new(|x| -x),
+    }
+}
+
+impl fmt::Display for PEResult {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use PEResult::*;
+        write!(
+            f,
+            "{}",
+            match self {
+                Evaluated(f) => f.to_string(),
+                Unevaluated(b) => b.to_string(),
+            }
+        )
+    }
+}
--- a/libslide/src/lib.rs
+++ b/libslide/src/lib.rs
@@ -1,2 +1,6 @@
-pub mod parser;
 pub mod scanner;
+
+pub mod parser;
+
+pub mod partial_evaluator;
+pub mod visitor;
--- a/libslide/src/parser.rs
+++ b/libslide/src/parser.rs
@@ -1,9 +1,14 @@
-mod types;
-use crate::scanner::{Token, TokenType};
+pub mod types;
+use crate::scanner::types::{Token, TokenType};
 use core::convert::TryFrom;
-pub use types::*;
+use types::*;
 
-pub struct Parser {
+pub fn parse(input: Vec<Token>) -> Box<Stmt> {
+    let mut parser = Parser::new(input);
+    parser.parse()
+}
+
+struct Parser {
     input: Vec<Token>,
     index: usize,
 }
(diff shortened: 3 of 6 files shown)


luke-bhan:
What do you mean? Like an alternative pattern?

ayazhafiz:
yeah

Отправлено: Monday, April 27, 2020 11:20:58 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/slide] Evaluate (#12)


What do you mean? Like an alternative pattern?

luke-bhan:
I mean Ayaz you know this there’s a million ways to evaluate this. I could write a recursive method or I could derive an iterator class. I was just thinking visitor is a very clean way to do it with good design principles. Maybe another route may be better

ayazhafiz:
Yes but either of those will become a visitor, no?

Отправлено: Monday, April 27, 2020 11:29:08 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/slide] Evaluate (#12)


I mean Ayaz you know this there’s a million ways to evaluate this. I could write a recursive method or I could derive an iterator class. I was just thinking visitor is a very clean way to do it with good design principles. Maybe another route may be better

ayazhafiz:
I will add some changes to this tomorrow, I hope I can show you that the visitor can work well.

Отправлено: Monday, April 27, 2020 11:29:08 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/slide] Evaluate (#12)


I mean Ayaz you know this there’s a million ways to evaluate this. I could write a recursive method or I could derive an iterator class. I was just thinking visitor is a very clean way to do it with good design principles. Maybe another route may be better

luke-bhan:
Ok. I just wouldn’t necessarily think you need a visitor is all - it’s just good practice

ayazhafiz:
Well, the data that we have is not homogenous. So anything you implement will end up looking like a visitor, even if you want to call it something else.

luke-bhan:
So you followed a similar approach with two levels. Also, theres some issues in modularity. We need to remove _ => unreachable!() correct? @ayazhafiz

ayazhafiz:
@luke-bhan I've implemented more visitor traits and completed the basic evaluator. Please take a look and let me know what you think.

There is now a bigger challenge to solve here -- in order to evaluate variables, we need to be able to return an evaluated type that may or may not be reduced. I think this could look something like `Result<UnevaluatedExpression, f64>`, but would probably need to be in a wrapper class to make processing easier. Basically what I am thinking is that we should be able to have the following:

```
1 + 2 => Result<f64>(3)
a + 2 => Result<UnevaluatedExpression>(BinaryOp(+ a 2))
```

This could get a little tricky because we will need to implement a strategy for optimistic reduction; e.g. we should be able to reduce `a + 1 + 2` to `a + 3`, but let's not worry about that yet.

Do you want to take a stab at implementing this data structure?

ayazhafiz:
> So you followed a similar approach with two levels. Also, theres some issues in modularity. We need to remove _ => unreachable!() correct? @ayazhafiz

Yes, I think the best way to do that is to define a new type of what can and can't be unary and binary operators. We can do that in another PR though.

luke-bhan:
So, do you want to do variables first?

ayazhafiz:
Yes. If you're up for it, do you want to take a shot at implementing this `Result<UnevaluatedExpression, f64>`-like type and adding support for partial evaluation of variables? In the meantime I can add unary/binary operator types and start working on the writer.

luke-bhan:
can't we just define a generic enum and declare that as type rather than build an entire class. For example

enum Result<T, E> {
    Ok(T),
    UnEval(E),
}

ayazhafiz:
Yes, that works. I'm just saying you might want to containerize this to
make evaluating a little bit easier. For example, how would you do
Result<T, E> + Result<T, E>? You probably want a helper for this.

luke-bhan:
I think this problem isn't that simple as defining helper methods. Like we could have situations where we don't want to add or we do, but its unclear. For example
(2^3)+b*a+(3+5)
goes to 
8+(b*a+8) which is going to be classified as an unevaluated expr
but in reality it should be 
b*8 + 16, but how do we know how to add to an unevaluated expr if precedence isn't clear . 
I think this problem will take me some thinking if we want to handle operators of these results that include the unevaluated part

ayazhafiz:
> but how do we know how to add to an unevaluated expr if precedence isn't clear

> I think this problem will take me some thinking if we want to handle operators of these results that include the unevaluated part

Yes, I agree -- that is a hard problem that will fall under the scope of the optimization platform. I think we shouldn't worry about that for now, and just do what we can with setting up the `PartialEvaluator`. Evaluating to

> 8+(b * a+8) which is going to be classified as an unevaluated expr

is fine for now.

luke-bhan:
Ok, so I added a PEResult type that handles evaluated and unevaluated. As of now, I just left unevaluated as a None since we haven't defined that implementation yet. I think that belongs in a different PR. Also, I added a dependency which is just a list of macros to shorten operator overloading. Should be very lightweight. lemme know what you think @ayazhafiz

ayazhafiz:
> Also, I added a dependency which is just a list of macros to shorten operator overloading. Should be very lightweight.

TBH, I would rather us not have any dependencies for this project (DIY kind beat). If needed we can implement `overload!` ourselves. What do you think?

libslide/Cargo.toml
```diff
@@ -5,3 +5,4 @@ authors = ["ayazhafiz <ayaz.hafiz.1@gmail.com>"]
 edition = "2018"
 
 [dependencies]
+overload = "0.1.1"
```
ayazhafiz, line 8: Also, I added a dependency which is just a list of macros to shorten operator overloading. Should be very lightweight.

TBH, I would rather us not have any dependencies for this project (DIY kind beat). If needed we can implement overload! ourselves. What do you think?

luke-bhan:
Had a feeling you’d say this. It’s just gonna be sloppier but shouldn’t be hard I’ll do it tomorrow.

ayazhafiz:
Sounds good. You may not even need a macro.

luke-bhan:
Ok, removed the dependency. LMK what you think @ayazhafiz

libslide/src/partial_evaluator/types.rs
```diff
@@ -0,0 +1,63 @@
+use std::ops;
+use crate::parser::types::Expr;
+
+pub struct PEResult {
+    pub unevaluated_expr: Option<Box<Expr>>,
+    pub evaluated: f64
+}
```
ayazhafiz, line 7: I think this would be better expressed as an enum like `Result` is:

```rust
pub enum PEResult {
  Evaluated(f64),
  Unevaluated(Box<Expr>),
}
```

this way when we need to add two `PEResult`s, we just do something like

```rust
impl ops::Add for PEResult {
  type Output = PEResult;

  fn add(self, rhs: PEResult) -> Self::Output {
    fold_binary(self, rhs, Operator::Plus)
  }
}

fn operator_table(op: Operator) -> Box<dyn Fn(f64, f64) -> f64 + 'static> {
  match op {
    Plus => Box::new(|x, y| x + y),
  }
}

fn fold_binary(lhs: PEResult, rhs: PEResult, op: Operator) -> PEResult {
  match lhs, rhs {
    Evaluated(x), Evaluated(y) => Evaluated(operator_table(op)(x, y)),
    Evaluated(x), Unevaluated(y) => Unevaluated(BinaryOperator{
      op,
      lhs: Number(x),
      rhs: y
    }),
    ...etc
  }
}
```

I hope this makes sense. One thing you may be wondering is why we would want to do dynamic allocation with a box for a function; I think this will not incur a runtime penalty because the `'static` lifetime is allocated for the lifetime of the program IIRC.

luke-bhan: why is this better?

ayazhafiz:
@luke-bhan ptal at the updates I made

luke-bhan:
Ok, So what exactly did you change because my changes are shown in your commit as well. I see you updated the visit variable function which looks good and added testing for unevaluated. Anything else I am missing?

ayazhafiz:
Nope, that's it. Also simplified the number of calls to evaluating an
expression.

luke-bhan:
Ok go ahead and merge!

ayazhafiz:
Okay. Do you want to go ahead and fix the issue with division precedence?

luke-bhan:
wait I am still having issues pulling

ayazhafiz:
No. Try this:

```bash
# make sure your branch is evaluate
git fetch origin
git reset --hard origin/evaluate
```

ayazhafiz:
Also can you approve this so I can merge

luke-bhan:
you are the reviewer blocking it

PR #14, opened 2020-04-29
Parse programs as a statement, refine operators, create parser macro

ayazhafiz:
There's a lot going on here, but I felt like doing it in one fell swoop
:slightly_smiling_face:. The big things are:

- Programs are now parsed as statements, which can either be an
  assignment or expression statement currently
- Operators on `BinaryExpr`s and `UnaryExpr`s are refined to be a
  variant of a small subset of `BinaryOperator`s and `UnaryOperator`s,
  respectively. This will with exhaustive eval matching in #12.
- Creates a macro for defining leveled binary operator parsers. This may
  be more complicated to understand, but simplifies the binary operator
  parsing code singificantly and should make it trivial to add new
  binary parsers later on if needed.

### DIFF
--- a/libslide/src/parser.rs
+++ b/libslide/src/parser.rs
@@ -1,16 +1,43 @@
 mod types;
 use crate::scanner::{Token, TokenType};
+use core::convert::TryFrom;
 pub use types::*;
 
-pub struct Parser<'a> {
-    input: &'a Vec<Token>,
+pub struct Parser {
+    input: Vec<Token>,
     index: usize,
 }
 
-impl<'a> Parser<'a> {
-    pub fn new(input: &Vec<Token>) -> Parser {
+macro_rules! binary_expr_parser {
+    ($self:ident $($name:ident: lhs=$lhs_term:ident, rhs=$rhs_term:ident, op=[$($matching_op:tt)+])*) => {
+        $(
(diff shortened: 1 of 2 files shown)


luke-bhan:
Wow this is alot. So now are we considering variable assignment to be its own operation. Also, I think having a binaryExpr and then a binaryOp is confusing.  also why are we making the parser a macro?

ayazhafiz:
> binaryExpr and then a binaryOp is confusing

Why is it confusing? One is an expression, one is an operator (function).

>  also why are we making the parser a macro?

The code for `add_sub_term`, `mul_div_mod_term`, `exp_term` parsers is very similar but there is a lot of it and it is somewhat confusing. It would be nice if we could centralize the code, which is what a macro allows for. I feel like the macro makes it more intuitive to understand how we are currently parsing a kind of binary expression (we make the LHS/RHS go up one level of precedence higher), and it will make it trivial to add a new binary expression parser without having to copy/paste a bunch of boilerplate.

ayazhafiz:
@luke-bhan do you want me to break this PR down into a few separate ones?

luke-bhan:
@ayazhafiz no I think this is good refactoring. Sorry for the delay in the response. Review should be accepted and ready to merge!

PR #23, opened 2020-04-30
Parser update

luke-bhan:
Added parsing from the left. Sorry it took me a while to solve this one. I only ended up adding a few lines and expanding all the macros. I think it can be cleaned up macro wise but I am still do not understand them well so if you want to go for it. Had to change tests in partial_evaluator as well since we left parse now.  @ayazhafiz

### DIFF
--- a/libslide/src/interpreter.rs
+++ b/libslide/src/interpreter.rs
@@ -1,0 +1,90 @@
+use crate::parser::types::*;
+use crate::scanner::TokenType;
+
+trait Data {
+    fn accept<V: Visitor>(&self, visitor: &mut V) -> V::Result;
+}
+
+trait Visitor {
+    type Result;
+    fn visit_binop(&mut self, item: &BinOp) -> Self::Result;
+    fn visit_unaryop(&mut self, item: &UnaryOp) -> Self::Result;
+    fn visit_float(&mut self, item: &f64) -> Self::Result;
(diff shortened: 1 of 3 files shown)


ayazhafiz:
Here is a simpler way to implement binary expressions you may want to consider: https://github.com/ayazhafiz/TypeScript/blob/95cc1c279ef8531b88f0a770ffad0b43ad9249d3/src/compiler/parser.ts#L4031

luke-bhan:
the above implementation is the exact same thing, just broken into separate parts. I will try and make this a more concise macro later today so its all one part

ayazhafiz:
Cool sounds good.

ayazhafiz:
@luke-bhan are you still planning to update this PR?

luke-bhan:
yes, but I haven't started on it if you wanted to do osmething else

ayazhafiz:
You're good. Let me know

luke-bhan:
@ayazhafiz PTAL, also I don't know how to combine this to one commit.

luke-bhan:
Theres an issue with the squashed script @ayazhafiz. Take a look at git branch --show-current. It does not exist as a command

ayazhafiz:
You need at least git 2.22:
https://stackoverflow.com/questions/1417957/show-just-the-current-branch-in-git/55088865#55088865

PR #28, opened 2020-05-03
Clean up public API and add slide app

ayazhafiz:


ayazhafiz:
> Approved, but we need to write language rules down for people wanting their stuff parsed which I feel like should be apart of this commit. I can add it if you agree @ayazhafiz

Sounds good, just submit a separate PR for it/

PR #31, opened 2020-05-05
Bignum

luke-bhan:
Create a big num class and refactor scanner and parser. I just added my initial work, will do more tomorrow!

### DIFF
--- a/libslide/src/bignum.rs
+++ b/libslide/src/bignum.rs
@@ -1,0 +1,114 @@
+use std::fmt;
+mod add;
+
+// this probably can go in utils but I put it in here for now
+trait StringUtils {
+    fn substring(&self, start: usize, len: usize) -> Self;
+}
+
+impl StringUtils for String {
+    fn substring(&self, start: usize, len: usize) -> Self {
+        self.chars().skip(start).take(len).collect()
+    }
+}
+
+fn to_u8(c: char) -> u8 {
+    c.to_digit(10).unwrap() as u8
+}
+
+fn to_char(n: u8) -> char {
+    std::char::from_digit(n as u32, 10).unwrap()
+}
+
+pub struct Bignum {
+    is_neg: bool,
+    // we use vector since it will allow faster computation time
+    // for smaller numbers
+    int: Vec<u8>,
+    dec: Vec<u8>,
+}
+
+impl Bignum {
+    pub fn new(input: String) -> Bignum {
+        let mut chars = input.chars();
+        let mut is_neg: bool = false;
+        let mut has_decimal = false;
+        let mut int = Vec::new();
+        let mut dec = Vec::new();
+        match chars.next() {
+            Some('-') => is_neg = true,
+            Some('.') => has_decimal = true,
+            Some(x) => int.push(to_u8(x)),
+            _ => (),
+        };
+        for c in chars {
+            if c.is_digit(10) {
+                if has_decimal {
+                    dec.push(to_u8(c));
+                } else {
+                    int.push(to_u8(c));
+                }
+            } else if c == '.' {
+                has_decimal = !has_decimal;
+                if !has_decimal {
+                    panic!("Number cannot have two decimal values");
+                }
+            } else {
+                panic!("Cannot parse string as number");
+            }
+        }
+        Bignum {
+            is_neg,
+            int: int.iter().rev().cloned().collect(),
+            dec,
+        }
+    }
+}
+
+impl fmt::Display for Bignum {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut result = String::new();
+        if self.is_neg {
+            result.push('-');
+        }
+        for n in self.int.iter().rev() {
+            result.push(to_char(*n));
+        }
+        if !self.dec.is_empty() {
+            result.push('.');
+            for n in self.dec.iter() {
+                result.push(to_char(*n));
+            }
+        }
+        write!(f, "{}", result)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    macro_rules! bignum_test {
+        ($($name:ident: $program:expr)*) => {
+        $(
+            #[test]
+            fn $name() {
+                use crate::bignum::Bignum;
+                let result = $program.to_string();
+                let bg = Bignum::new(result.clone());
+                assert_eq!(bg.to_string(), result);
+            }
+        )*
+        }
+    }
+    mod num {
+        bignum_test! {
+            int: "5"
+            negative_int: "-5"
+            float: "5.5"
+            negative_float: "-5.5"
+            bigger_int: "1002039444884993020"
+            bigger_float: "102399959939999.393993"
+            decimal: ".3"
+            negative_decimal: "-.3"
+        }
+    }
+}
--- a/libslide/src/bignum/add.rs
+++ b/libslide/src/bignum/add.rs
@@ -1,0 +1,119 @@
+#![allow(clippy::suspicious_arithmetic_impl)]
+use crate::bignum::Bignum;
+use std::mem;
+use std::ops;
+
+impl ops::Add for Bignum {
+    type Output = Bignum;
+
+    fn add(self, rhs: Bignum) -> Bignum {
+        let is_neg: bool = rhs.is_neg && self.is_neg;
+
+        // this assumes the signs are both positive for now
+        let mut carry: u8 = 0;
+        let mut lhs_size: usize = self.dec.len();
+        let mut rhs_size: usize = rhs.dec.len();
+        let mut lhs_vec: Vec<u8>;
+
+        // make lhs smaller vector
+        // lhs_size = self, rhs_self = rhs
+        let mut res_dec = if lhs_size > rhs_size {
+            mem::swap(&mut lhs_size, &mut rhs_size);
+            lhs_vec = rhs.dec;
+            self.dec
+        } else {
+            lhs_vec = self.dec;
+            rhs.dec
+        };
+
+        for i in (0..lhs_size).rev() {
+            res_dec[i] += lhs_vec[i] + carry;
+            carry = (res_dec[i] > 9) as u8;
+            res_dec[i] %= 10;
+        }
+
+        // 2. Handle Integer
+
+        lhs_size = self.int.len();
+        rhs_size = rhs.int.len();
+
+        // make lhs smaller vector
+        let mut res_int = if lhs_size > rhs_size {
+            mem::swap(&mut lhs_size, &mut rhs_size);
+            lhs_vec = rhs.int;
+            self.int
+        } else {
+            lhs_vec = self.int;
+            rhs.int
+        };
+
+        // compute addition with both ints
+        for i in 0..lhs_size {
+            res_int[i] += lhs_vec[i] + carry;
+            carry = (res_int[i] > 9) as u8;
+            res_int[i] %= 10;
+        }
+
+        // fill with rest of the larger integer vector
+        // note we need to propogate carry here while in decimal we do not, but if carry = 0 we are
+        // done since res_int already contains the values
+        let mut i = lhs_size;
+        while carry != 0 && i < rhs_size {
+            res_int[i] += carry;
+            carry = (res_int[i] > 9) as u8;
+            res_int[i] %= 10;
+            i += 1;
+        }
+
+        // add 1 if a carry is leftover
+        if carry == 1 {
+            res_int.push(1);
+        }
+
+        Bignum {
+            is_neg,
+            int: res_int,
+            dec: res_dec,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    macro_rules! bignum_test_add {
+        ($($name: ident: $lhs:expr, $rhs:expr, $program:expr)*) => {
+        $(
+            #[test]
+            fn $name() {
+                use crate::bignum::Bignum;
+                let lhs = Bignum::new($lhs.to_string());
+                let rhs = Bignum::new($rhs.to_string());
+                let result = $program.to_string();
+                assert_eq!((lhs+rhs).to_string(), result);
+            }
+        )*
+        }
+    }
+
+    mod add {
+        bignum_test_add! {
+            int1: "5", "5", "10"
+            int2: "55", "13", "68"
+            int3: "5", "15", "20"
+            int4: "55555555555555555555555", "5555555555555555", "55555561111111111111110"
+            int5: "99999999999999999999999999999999" , "999999999999999999999", "100000000000999999999999999999998"
+            int6: "111", "1", "112"
+            float: "0.1", "0.1", "0.2"
+            float2: "0.111", "0.1", "0.211"
+            float3: "0.1", "0.111", "0.211"
+            float4: "0.5", "0.9", "1.4"
+            float5: "0.55555", "0.99", "1.54555"
+            float6: "0.555555555555555555555555555", "0.555555555555555555", "1.111111111111111110555555555"
+            float7: "0.99999999999999999", "0.999999999999999999999999999999", "1.999999999999999989999999999999"
+            float8: "0.1112", "0.923", "1.0342"
+            mixed: "1.1", "1.1", "2.2"
+            mixed2: "1.9999999", "9999.99999", "10001.9999899"
+            mixed3: "99999999999999.999999999999", "99999999999.99999", "100099999999999.999989999999"
+        }
+    }
+}
--- a/libslide/src/lib.rs
+++ b/libslide/src/lib.rs
@@ -1,3 +1,6 @@
+mod bignum;
+pub use bignum::Bignum;
+
 mod scanner;
 pub use scanner::scan;
 pub use scanner::ScannerOptions;


libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
+use std::fmt;
+const DIGIT_COUNT: usize = 10000;
```
ayazhafiz, line 2: It would be cheaper to use a `Vec` IMO. I think Rust/LLVM will stack allocate small-capacity vecs.

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+// this probably can go in utils but I put it in here for now
+trait StringUtils {
+    fn substring(&self, start: usize, len: usize) -> Self;
+}
+
+impl StringUtils for String {
+    fn substring(&self, start: usize, len: usize) -> Self {
+        self.chars().skip(start).take(len).collect()
+    }
+}
```
ayazhafiz, line 13: 👍 

I wonder if you could do something like

```suggestion
impl StringUtils for String {
    fn substring(&self, start: usize, len: usize) -> Self {
        self[start..start+len].to_owned()
    }
}
```

?

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub fn num(input: String) -> Bignum {
+    let bg = Bignum::new(input);
+    bg
+}
```
ayazhafiz, line 18: It's probably better to just expose `BigNum` and make its struct fields private.

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+pub struct Bignum {
+    // False is positive, True is negative
+    pub sign: bool,
+    pub arr: [char; DIGIT_COUNT],
+    pub size: usize,
+}
+
+impl Bignum {
+    pub fn new(input: String) -> Bignum {
+        let mut arr_val: [char; DIGIT_COUNT] = ['0'; DIGIT_COUNT];
```
ayazhafiz, line 30:
```suggestion
        let mut arr: [char; DIGIT_COUNT] = ['0'; DIGIT_COUNT];
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+        let sign_val: bool;
```
ayazhafiz, line 31:
```suggestion
        let sign: bool = false;
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+        let s : String;
+        let mut size_val: usize = 0;
```
ayazhafiz, line 33:
```suggestion
        let mut size: usize = 0;
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub fn num(input: String) -> Bignum {
+    let bg = Bignum::new(input);
+    bg
+}
+
+
+pub struct Bignum {
+    // False is positive, True is negative
+    pub sign: bool,
```
ayazhafiz, line 23: why not call it `neg` then?

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub arr: [char; DIGIT_COUNT],
+    pub size: usize,
+}
+
+impl Bignum {
+    pub fn new(input: String) -> Bignum {
+        let mut arr_val: [char; DIGIT_COUNT] = ['0'; DIGIT_COUNT];
+        let sign_val: bool;
+        let s : String;
```
ayazhafiz, line 32:
```suggestion
        let mut s = input.into_iter()
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+        let mut size_val: usize = 0;
+        if input.chars().next().unwrap() == '-' {
+            sign_val = true;
+            s = input.substring(1, input.chars().count()-1);
```
ayazhafiz, line 36:
```suggestion
            s.next(); // if you change it to an iterator
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+        else {
+            sign_val = false;
+            s = input;
+        }
```
ayazhafiz, line 41:
```suggestion
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,100 @@
...
+        for (i, c) in s.chars().enumerate() {
```
ayazhafiz, line 42:
```suggestion
        for (i, c) in s { // if you change s to an iterator
```

libslide/src/bignum.rs
luke-bhan: and just build the vec dynamically every time a new object is made?

luke-bhan: Do you really think its that hard? I found some c++ files where its not that bad?

luke-bhan: And just build the vec dynamically at runtime?

libslide/src/bignum.rs
luke-bhan: no no no. Rust does not allow subscript operator for strings.

libslide/src/bignum.rs
luke-bhan: gotcha will change!

libslide/src/bignum.rs
luke-bhan: will change

libslide/src/bignum.rs
ayazhafiz: Do they implement division and stuff? I’ve heard it’s pretty challenging. But I’d love to see it done!

ayazhafiz: Yeah, I’ll let you make the choice on this one. Either dynamic allocation with smaller sizes or static allocation of a large number of bits. Each have their pros and cons, I’ll defer to you.

libslide/src/bignum.rs
ayazhafiz: What about slices? https://doc.rust-lang.org/book/ch04-03-slices.html#string-slices

libslide/src/bignum.rs
luke-bhan: Division is just repeated subtraction though. Once you implement that it’s done. I think the tricky think would be exp and overflows

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "scripts/_scripts"]
 	path = scripts/_scripts
 	url = https://github.com/ayazhafiz/scripts
+[submodule "scripts/scripts"]
```
ayazhafiz, line 4: Please remove. Use git submodule update —recursive if you need to install it

libslide/src/bignum.rs
luke-bhan: They don’t work Bc slices are byte based and we need character based.

luke-bhan: They don’t work Bc slices are byte based and we need character based.

ayazhafiz: ahh good call

libslide/src/bignum.rs
ayazhafiz: Please fix

libslide/src/bignum.rs
```diff
@@ -0,0 +1,107 @@
...
+    bg
+}
+
+pub struct Bignum {
+    // False is positive, True is negative
+    neg_sign: bool,
+    // we use vector since it will allow faster computation time
+    // for smaller numbers
+    int: Vec<u32>,
+    dec: Vec<u32>,
```
ayazhafiz, line 26: No need for a `u32` since we're just holding digits. u8 will do

libslide/src/bignum.rs
```diff
@@ -0,0 +1,107 @@
...
+        let s: String;
+        let mut int: Vec<u32> = Vec::new();
+        let mut dec: Vec<u32> = Vec::new();
+        if input.chars().next().unwrap() == '-' {
+            neg_sign = true;
+            s = input.substring(1, input.chars().count() - 1);
+        } else {
+            neg_sign = false;
+            s = input;
+        }
```
ayazhafiz, line 41:
```suggestion
        let neg_sign: bool = false;
        let mut s = input.into_iter();
        let mut int: Vec<u32> = Vec::new();
        let mut dec: Vec<u32> = Vec::new();
        if let Some('-') = s.next() {
            neg_sign = true;
        }
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,107 @@
...
+        let mut is_decimal = false;
```
ayazhafiz, line 42:
```suggestion
        let mut has_decimal = false;
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,107 @@
...
+        for c in s.chars() {
+            match c {
+                x if x.is_digit(10) && is_decimal == false => int.push(c.to_digit(10).unwrap()),
+                x if x.is_digit(10) => dec.push(c.to_digit(10).unwrap()),
+                x if x == '.' => match is_decimal {
+                    true => panic!("string cannot have two decimal points!"),
+                    false => is_decimal = true,
+                },
+                _ => panic!("String contains not decimal values"),
+            }
```
ayazhafiz, line 52:
```suggestion
            if c.is_digit(10) {
                int.push(c.to_digit(10).unwrap());
            } else if c == '.' {
                match is_decimal {
                    true => panic!("string cannot have two decimal values"),
                    false => is_decimal = true,
            } else {
                panic!("cannot parse string as number")
            }
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+        Bignum {
+            neg_sign: neg_sign,
+            int: int,
+            dec: dec,
```
ayazhafiz, line 57:
```suggestion
            neg_sign,
            int,
            dec,
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl fmt::Display for Bignum {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut result: String = "".to_string();
+        if self.neg_sign {
+            result.push('-');
+        }
+        for i in 0..self.int.iter().count() {
+            result.push(from_digit(self.int[i], 10).unwrap());
+        }
```
ayazhafiz, line 70:
```suggestion
        for i in 0..self.int.iter() {
            result.push(i.into());
        }
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,107 @@
...
+        if self.dec.iter().count() != 0 {
```
ayazhafiz, line 71: This consumes the entire iterator. It's better to do something like
```suggestion
        if self.dec.len() > 0 {
```

libslide/src/bignum.rs
```diff
@@ -0,0 +1,107 @@
...
+            result.push('.');
+            for i in 0..self.dec.iter().count() {
+                result.push(from_digit(self.dec[i], 10).unwrap());
+            }
```
ayazhafiz, line 75:
```suggestion
            for i in 0..self.dec.iter() {
                result.push(i.into());
            }
```

libslide/src/bignum/bignum_add.rs
```diff
@@ -0,0 +1,143 @@
+mod bignum_add {
```
ayazhafiz, line 1: This doesn't need to be wrapped in a mod. Also, just call the file `add.rs`

libslide/src/bignum/bignum_add.rs
```diff
@@ -0,0 +1,143 @@
...
+    use crate::bignum::Bignum;
+    use std::ops;
+
+    // @todo - figure out a swap method to handle lines 17-26
+    impl ops::Add for Bignum {
+        type Output = Bignum;
+
+        fn add(self, rhs: Bignum) -> Bignum {
+            // this assumes the signs are both positive for now
```
ayazhafiz, line 10: If they are negative, just do (-1 * (-self + -rhs)). If one is negative and one is positive use subtraction.

libslide/src/bignum/bignum_add.rs
```diff
@@ -0,0 +1,143 @@
...
+            let mut carry: u8 = 0;
+            let mut lhs_size: usize = self.dec.iter().count();
+            let mut rhs_size: usize = rhs.dec.iter().count();
+            let mut res_dec: Vec<u8> = Vec::new();
+            let mut big_vec: Vec<u8>;
```
ayazhafiz, line 15: Just make one vector to hold the resulting decimal value. Initially make it the bigger decimal, and you can allocate more if needed at the end.

libslide/src/bignum/bignum_add.rs
```diff
@@ -0,0 +1,143 @@
...
+
+            // swap if neccessary to make lhs smaller. Initialize big vec as the bigger vector
+            if lhs_size > rhs_size {
+                let tmp: usize = lhs_size;
+                lhs_size = rhs_size;
+                rhs_size = tmp;
+                big_vec = self.dec.clone();
+            } else {
+                big_vec = rhs.dec.clone();
+            }
```
ayazhafiz, line 27: You could use mem::swap (import std::mem):
```suggestion
            if lhs_size > rhs_size {
                mem::swap(&lhs_size, rhs_size);
                big_vec = self.dec.clone();
            } else {
                big_vec = rhs.dec.clone();
            }
```

libslide/src/bignum/bignum_add.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        fn add(self, rhs: Bignum) -> Bignum {
+            // this assumes the signs are both positive for now
+            let mut carry: u8 = 0;
+            let mut lhs_size: usize = self.dec.iter().count();
+            let mut rhs_size: usize = rhs.dec.iter().count();
+            let mut res_dec: Vec<u8> = Vec::new();
+            let mut big_vec: Vec<u8>;
+
+            // HANDLE DECIMAL
```
ayazhafiz, line 17:
```suggestion
            // 1. Handle decimal. len(lhs.dec) < len(rhs.dec)
```

libslide/src/bignum/bignum_add.rs
```diff
@@ -0,0 +1,143 @@
...
+            res_dec = res_dec.iter().rev().cloned().collect();
+
+            // add the remaining decimals of the larger vector
+            let mut i = lhs_size;
+            while i < rhs_size {
+                res_dec.push((big_vec[i]) % 10);
+                i += 1;
+            }
+
+            // HANDLE INTEGER
```
ayazhafiz, line 49:
```suggestion
            // 2. Handle integer
```

libslide/src/bignum/bignum_add.rs
```diff
@@ -0,0 +1,143 @@
...
+    use std::ops;
+
+    // @todo - figure out a swap method to handle lines 17-26
+    impl ops::Add for Bignum {
+        type Output = Bignum;
+
+        fn add(self, rhs: Bignum) -> Bignum {
+            // this assumes the signs are both positive for now
+            let mut carry: u8 = 0;
+            let mut lhs_size: usize = self.dec.iter().count();
```
ayazhafiz, line 12: What? just use `len()`. This is O(n), len is O(1)

libslide/src/bignum/bignum_add.rs
```diff
@@ -0,0 +1,143 @@
...
+                lhs_size = rhs_size;
+                rhs_size = tmp;
+                big_vec = self.dec.clone();
+            } else {
+                big_vec = rhs.dec.clone();
+            }
+
+            // compute the decimal addition up to the size of the small vec
+            for i in (0..lhs_size).rev() {
+                res_dec.push((self.dec[i] + rhs.dec[i] + carry) % 10);
```
ayazhafiz, line 31: This is assuming self is the normalized `lhs`. What you should do is get a reference to both and swap the references if they should be:

```rust
let mut lhs_dec = &&self.dec; // why double reference? We want a pointer to the reference of the vector, so we can mutate what we're pointing at.
let mut rhs_dec = &&rhs.dec;

if (*lhs_dec).len() > (*rhs_dec).len() {
    std::mem::swap(&mut lhs_dec, &mut rhs_dec);
}
```

ayazhafiz, line 31: better to allocate a u8 (maybe `res`) outside this loop, then do

```rust
res = self.dec[i] + rhs.dec[i] + carry;
res_dec[i] = res % 10; // make res_dec the larger decimal to start!
carry = (res >9) as u8; // saves a branch, in rust true==1, false==0
```

libslide/src/bignum/bignum_add.rs
luke-bhan: yep thats the plan

libslide/src/bignum/bignum_add.rs
luke-bhan: see I read this was slow is the only thing so I wasn't sure

ayazhafiz:
Where did you read that? It's just a memswap

libslide/src/bignum/bignum_add.rs
luke-bhan: wait what? Big_Vec is need to store the larger vector of lhs and rhs?

libslide/src/bignum/bignum_add.rs
luke-bhan: why is this any better? you're calling .len() everytime you need the size.

ayazhafiz: Oh I'm dumb. nevermind

libslide/src/bignum/add.rs
```diff
@@ -0,0 +1,135 @@
...
+            self.dec.clone()
+        } else {
+            rhs.dec.clone()
+        };
+
+        // compute the decimal addition up to the size of the small vec
+        let mut res: u8;
+        for i in (0..lhs_size).rev() {
+            res = self.dec[i] + rhs.dec[i] + carry;
+            res_dec.push(res % 10);
```
ayazhafiz, line 32: I don't think we should be pushing on an empty vec. There is a really easy way to solve this and the rest of your code on L36-44, which is to just initialize res_dec as the larger of the two vectors (self.dec or rhs.dec).

libslide/src/bignum/add.rs
```diff
@@ -0,0 +1,135 @@
...
+            i += 1;
+        }
+
+        // 2. Handle Integer
+
+        lhs_size = self.int.len();
+        rhs_size = rhs.int.len();
+
+        // we reverse the vectors since our carry matters when we handle
+        // adding the larger vector
```
ayazhafiz, line 52: Maybe we should store the integer part in the BigNum with the LSB at the 0 index. Otherwise we are reversing twice each time we do any operation, which is way more expensive than reversing once for formatting (which is expensive anyway)

libslide/src/bignum/add.rs
```diff
@@ -0,0 +1,135 @@
...
+        let mut res_int: Vec<u8> = Vec::new();
```
ayazhafiz, line 53: No need to create a new vector with no preallocation, just make it the larger of the two existing vectors.

libslide/src/bignum/add.rs
luke-bhan: these comments are wrong and should've been deleted on my latest refactor. I rewrote the algo so it doesn't need to reverse the initial vectors. It should be the min time complexity of O(lhs+rhs)

libslide/src/bignum/add.rs
luke-bhan: You need to make this one a new vector because you need the push_back method for the second part of this logic since it is easier to not store an i. I did switch it for the dec part.

libslide/src/bignum/add.rs
ayazhafiz: Still, we reverse the vector at the end.

libslide/src/bignum/add.rs
luke-bhan: What is wrong with pushing an empty vec?

libslide/src/bignum/add.rs
luke-bhan: even if it was built the opposite way, then we would have to reverse the dec vector at the end

libslide/src/bignum/add.rs
ayazhafiz: Way more allocations

libslide/src/bignum/add.rs
ayazhafiz: Why?

ayazhafiz: You wouldn't have to reverse it at all

ayazhafiz: If you keep 123.456 as
int: 321
dec: 456

libslide/src/bignum/add.rs
ayazhafiz: I'm not sure I understand that. Why do you need push_back?

libslide/src/bignum/add.rs
luke-bhan: I see.Let me finish subtract and ill refactor.

libslide/src/bignum/add.rs
luke-bhan: I think I will rewrite this so I can solve a few of these issues. I don't think we need push_back.

ayazhafiz: Sg

luke-bhan:
@ayazhafiz updated the add method to not use allocations. Lmk what you think

libslide/src/bignum/add.rs
```diff
@@ -0,0 +1,134 @@
...
+            lhs_vec = self.dec;
+            rhs_vec = rhs.dec;
+        }
+        res_dec = rhs_vec.clone(); 
+        let mut i = rhs_size as i8 - 1;
+        // add decimals that are unmatched by smaller vec
+        while i >= lhs_size as i8 {
+            res_dec[i as usize] = rhs_vec[i as usize];
+            i -= 1;
+        }
```
ayazhafiz, line 39: res_dec is a clone of rhs_vec, so there is no need to do this

libslide/src/bignum/add.rs
```diff
@@ -0,0 +1,134 @@
...
+        let mut lhs_vec: Vec<u8>;
+        let mut rhs_vec: Vec<u8>;
+        let mut res: u8;
+
+        // 1. Handle Decimal
+
+        let mut res_dec: Vec<u8>;
+
+        // make lhs smaller vector
+        if lhs_size > rhs_size {
```
ayazhafiz, line 25: I think it might be easier to understand if you compare self.dec and rhs.dec sizes here, then set lhs_size and rhs_size inside the conditional.

libslide/src/bignum/add.rs
```diff
@@ -0,0 +1,134 @@
...
+            res = rhs_vec[i] + carry;
+            println!("{}", res % 10);
+            res_int[i] = res % 10;
+            carry = (res > 9) as u8;
+        }
+
+        // add 1 if a carry is leftover
+        if carry == 1 {
+            res_int.insert(0, 1);
+        }
```
ayazhafiz, line 86: Let's just push_back and flip how the integer is stored in the `BigNum`. insert is expensive because we have to shift all the elements.

luke-bhan: We can't flip how the integer is stored. We don't know where the decimal is located so to build from the decimal out we would have to flip it anyway.

luke-bhan: It would only slow performance because we would need to do reverses for both BigNums rather then just one shift here

ayazhafiz: We only parse the number once, so flipping the decimal during construction is okay (just one O(n)) operation. But having to call insert everytime we do addition is way more expensive, because we now have an O(n) operation * # of times we call add. Plus, flipping is actually really only O(n/2) instructions.

Does this make sense? Basically what I mean is that for a number 123.456 we should store the int as 321. One reversal will be required during parsing and no reversals will be required during addition, at most one insertion to the end of the vector will be required.

libslide/src/bignum/add.rs
luke-bhan: Yes, it would, but we would loose performance because we would need to calculate self.dec.len() twice. One here and once in the swap

libslide/src/bignum/add.rs
luke-bhan: done.

libslide/src/bignum/add.rs
```diff
@@ -0,0 +1,129 @@
...
+            mem::swap(&mut lhs_size, &mut rhs_size);
+            lhs_vec = rhs.int;
+            rhs_vec = self.int;
+        } else {
+            lhs_vec = self.int;
+            rhs_vec = rhs.int;
+        }
+        res_int = rhs_vec.clone();
+
+        // computer intial addition
```
ayazhafiz, line 60:
```suggestion
        // compute addition with both ints
```

libslide/src/bignum/add.rs
```diff
@@ -0,0 +1,129 @@
...
+        for i in 0..lhs_size {
+            res = rhs_vec[i] + lhs_vec[i] + carry;
+            res_int[i] = res % 10;
+            carry = (res > 9) as u8;
```
ayazhafiz, line 64: you can even get rid of res:

```suggestion
            res_int[i] += lhs_vec[i] + carry;
            carry = (res_int[i] > 9) as u8;
            res_int[i] %= 10;
```

luke-bhan:
I think I’m just going to do the whole class in this PR or would you prefer separate PR for every op @ayazhafiz

ayazhafiz:
Let's break it up into a few separate PRs

ayazhafiz:
Don't worry about the linter errors, I'll fix that up rn

Issue #7, opened 2020-04-24
Optimization platform


ayazhafiz:
We should begin work on the expression optimization platform now that we have an AST built. Here is the rough architecture we had originally designed for slide:

![60360630726__F5CA088E-3C3E-4BE6-860C-13B44BAA3F87](https://user-images.githubusercontent.com/20735482/80169413-2dea2980-859a-11ea-9a92-cc6691ac4e5c.jpg)

I think we need to at least think about the following things as we go forward:

- [ ] Do we need a normalizer to express expressions in a common format? This would allow us to rearrange expressions safely when they can be to "align" common items like variables, e.g. `(a + b) - (b + a) -> (a + b) - (a + b)`. Maybe this will make some optimizations easier, but I am not convinced of it.
- [ ] How do we design a platform where we can easily plug optimizations in and out? We will need to do this as a some kind of rule-based registry, but what does that actually look like?
- [ ] What algorithms do we use to evaluate an expression and determine whether it can be optimized? Should this algorithm be defined as part of a rule, or should a rule be agnostic to the algorithm we use to evaluate a rule?

ayazhafiz:
@luke-bhan do you want to hop on a call sometime soon to chat about this?

luke-bhan:
im free now

ayazhafiz:
I'm not. What about this weekend?

ayazhafiz:
text me xoxo gossip girl 💋

luke-bhan:
im free pretty much anytime lol

ayazhafiz:
@luke-bhan I think one optimization we can start with is associativity. If an expression is unevaluatable, we check if it can be rewritten in an associative manner and recompute it. Does that make sense?

luke-bhan:
yes, so if we have an issue like 3+a+4, we will check if a+3+4 is computable

luke-bhan:
Do you agree we can consider an optimization to be smaller by the number of ops it has? That’s the way we can tell?

luke-bhan:
Also this gonna make the program run a lot slower Bc we’re gonna have need tokens Combination available switches like tCa

ayazhafiz:
> yes, so if we have an issue like 3+a+4, we will check if a+3+4 is computable

Well that is commutativity, but yeah. Idea stays.

> Do you agree we can consider an optimization to be smaller by the number of ops it has? That’s the way we can tell?

Yeah that seems reasonable.

[...]

PR #33, opened 2020-05-05
Consolidate Float and Int expressions as Const

ayazhafiz:
https://github.com/ayazhafiz/slide/issues/7#issuecomment-623786381


luke-bhan:
I think we should change our tokentype here to just hold float @ayazhafiz

ayazhafiz:
good call, i'll fix

ayazhafiz:
@luke-bhan ptal

PR #34, opened 2020-05-06
Expression term-rewritting rule platform

ayazhafiz:
This commit starts implementation of the term-rewritting rule system.
See #7 for more details on the implementation.

I'll try to make the code pretty for the long-term arch in this PR, but
will clean up/add more comments later on.


libslide/src/partial_evaluator/rules.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+impl From<&String> for BuiltRule {
+    /// Converts a string representation of a rule to a `BuiltRule`.
+    /// A rule must be expressed in the string form
+    ///   <expr> -> <expr>
+    /// Where <expr> is an expression.
+    fn from(rule: &String) -> Self {
+        fn is_var_char(c: char) -> bool {
+            return c.is_alphabetic() || c == '$' || c == '#' || c == '_';
```
luke-bhan, line 46: why do we have $ # and _ here?

luke-bhan:
looks good so far. From what I can see we have it so that we are reading in rules, parsing them, and adding that to a list pretty much. Am I missing anything ? @ayazhafiz

ayazhafiz:
> looks good so far. From what I can see we have it so that we are reading in rules, parsing them, and adding that to a list pretty much. Am I missing anything ? @ayazhafiz

Nope, that's it :)

libslide/src/partial_evaluator/rules.rs
ayazhafiz: See the grammar in the linked issue. Later we will use # and _ for other variables.

ayazhafiz:
@luke-bhan this should now be ready for review! Our app now supports one simplification: `$a + 0 -> $a`, but more to come after this rule matching system lands :)

The PR is very large because I included a refactoring of some files as well and removed some stale ones (like partial evaluator code); sorry about that. If it's too confusing let me know and I can try to break this PR down now.

Let me know what you think.

luke-bhan:
looking through now @ayazhafiz

luke-bhan:
you need to add comments my friend! @ayazhafiz It is pretty tricky to understand this entire pr when there is only two in the pattern file.

ayazhafiz:
Lol. Okay, should I add comments and re-request review?

luke-bhan:
yes. And make them super dumbed down. We can remove them before merging it just will help with my understanding :).

ayazhafiz:
Okay I'll do it soon

ayazhafiz:
@luke-bhan added a _lot_ of comments, ptal :)

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -16,8 +24,8 @@ pub fn match_rule(rule: Expr, target: Expr) -> Option<Replacements> {
             }
```
luke-bhan, line 24: This can be simplified to 
if a = b {
Some(Replacements::default())
}
none

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -102,6 +119,7 @@ impl Replacements {
                 }
```
luke-bhan, line 119: We shouldn't just return none. I think a nice way to do it is return a Result < self, self::error> that way we mirror the try_from logic rust already mandates.

libslide/src/evaluator_rules.rs
```diff
@@ -17,11 +17,40 @@ static DEFAULT_RULESET: &[&str] = &[
...
+/// expression, differing only in its pattern matching variables. The form of pattern matching
+/// variables and the expressions they match are as follows:
+///   
+///   | pattern | matches        |
+///   |:------- |:-------------- |
+///   | _<name> | Any expression |
+///   | #<name> | A constant     |
+///   | $<name> | A variable     |
+///   | <name>  | A variable     |
+///
```
luke-bhan, line 36: why do we have $ if its a variable even without it? line 17 in variables.rs implies the $ isn't necessary. I may be missing a reasoning for this.

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: Why not? What would the error be? To me, an Option equivalent to  `Result<Self, ()>` where the error is implicit - processing failed.

libslide/src/evaluator_rules.rs
ayazhafiz: I would like for just `$name` to be able to parse a variable. I'll change it, though it will require some changes to our parser -- that is why I allow `name` to be a variable too, because by default the parser will parse it as a var as well.

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: fixed

libslide/src/evaluator_rules/pattern_matching.rs
luke-bhan: I think its better code to make errors explicit for anyone who looks or uses our library. Implicit errors can be impossible to find and theres certainly going to be bugs.

libslide/src/evaluator_rules.rs
luke-bhan: ok, dont' worry about it on this PR. Add a todo

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: In general I agree with you, but in this method we only fail in one place that is documented on the function definition. Furthermore, (1) this method is internal to libslide and will not be available publicly, and (2) no one would consume the result of the error even if we did provide it, because it is not propogated up if we fail to match a rule -- we just skip it and move on.

libslide/src/evaluator_rules.rs
ayazhafiz: Added

libslide/src/evaluator_rules/pattern_matching.rs
luke-bhan: will try_merge fail often then if we are matching multiple rules.

luke-bhan: I am not sure I understand the point of this method and how it will be used.

ayazhafiz: Yes, but separately each time. It is meant to merge two sets of replacements. Basically the idea is let's say I have a binary expression "1 + 2" that I want to match against "#a + #a". Matching each LHS of the target expression against each LHS of the rule will work, and I will get two sets of replacements: "{#a: 1}" and "{#a: 2}". The problem is, this is not actually valid across the entire addition expression, because 1 != 2. So when determining the replacements for the entire addition expression, we have to merge these two replacements sets, at which point we determine that they are invalid and therefore there is no matching pattern.

ayazhafiz:
Gn! I'll wait for your response on try_merge before merging.

libslide/src/evaluator_rules/pattern_matching.rs
luke-bhan: i got it. So we are basically applying rules where the terms may not neccesarily be the same. Thanks for explaining. we don't need an error.

luke-bhan:
go ahead and merge!

PR #37, opened 2020-05-07
Remove support for arbitrary variable matching variable pattern

ayazhafiz:
This also adds pattern tokens to the scanner and refactors the parser to
be a trait with two implementations -- an ExpressionPattern that errors
when it sees a pattern (which should not be in pure expressions), and an
ExpressionPatternParser (which specially parses patterns and errors on
variables).

This still isn't really pretty because we lose pattern information
during parsing, since all patterns are currently lowered to variables. I
think the way to fix this would be to introduce an extended grammar for
the expression pattern parser, but we can do that in a separate patch.


luke-bhan:
wait this pr and the prev one both affect the scanner and parser is that ok?

ayazhafiz:
Going with the iter

PR #40, opened 2020-05-08
Add commutative rule and pattern matching tests

ayazhafiz:
Continues expanding the optimization platform (#7)

### DIFF
--- a/libslide/src/evaluator_rules/pattern_matching.rs
+++ b/libslide/src/evaluator_rules/pattern_matching.rs
@@ -22,13 +22,30 @@ pub fn match_rule(rule: Expr, target: Expr) -> Option<Replacements> {
             // Constant values are... constant, so there is no need to replace them.
             Some(Replacements::default())
         }
-        (Var(var), var_e @ Var(_)) => {
-            if VariablePattern::from_name(&var.name) & VariablePattern::Variable == 0 {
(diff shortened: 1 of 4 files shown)


libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -138,42 +165,93 @@ mod tests {
...
+                    let repls = match_rule(parsed_rule, parsed_target);
+                    let (repls, expected_repls): (Replacements, Vec<&str>) =
+                        match (repls, $expected_repls) {
+                            (None, expected_matches) => {
+                                assert!(expected_matches.is_none());
+                                return;
+                            }
+                            (Some(repl), expected_matches) => {
+                                assert!(expected_matches.is_some());
+                                (repl, expected_matches.unwrap())
```
luke-bhan, line 200: won't unwrap through here if there is nothing so we don't need the assert?

ayazhafiz: This is just to make the error messages a little bit nicer. If it panics, we just get a stack trace and it's not super clear where the error happened. With the assertion we're told where the error took place.

luke-bhan: gotcha. smart.

PR #41, opened 2020-05-08
Do not clone int during bignum creation

ayazhafiz:
The vec is only used once here, so just move it into the iterator rather
than making a new allocation.


luke-bhan:
We neecd to change the add method as well

ayazhafiz:
This doesn't change the implementation, it's all the same

luke-bhan:
No, I understand that. I am saying I use the same cloning form in the add method and we should refactor that with this pr now too

ayazhafiz:
Do you? I didn’t see it

PR #42, opened 2020-05-09
Separate pattern and expression grammars

ayazhafiz:
This commit is a real big kahuna. It separates the grammar for raw
expressions from the grammar for expression patterns, thereby also
creating two separate parsers for expressions vs expression patterns.

What is the advantage of this? Mostly better typing. This change is a
large amount of code churn with duplication and possibly-excessive
typing in the grammar and parser, but I think in the long-term this is a
good design choice.

Consider, for example, the types in `evaluator_rules` and
`evaluator_rules/pattern_matching`. Previously, a pattern matching took
in two expressions and a set of replacements was between two
expressions. This means that a matching and replacement could occur
between any two arbitrary expressions, when really it can only occur
between an expression pattern and an expression. Though this distinction
is minute, it is an important semantic difference -- expressions and
expression patterns are distinct ideas, since the latter is a
generalization of the former and typically not used to represent what a
consumer may categorize as an expression. The semantic differentiation
means we can now say that `Replacements` are `ExprPat -> Expr`, that
`Replacements` implements a `Transformer<ExprPat, Expr>` (meaning it
transforms `ExprPat`s to `Expr`s), and that a `BuiltRule` implements a
`Transformer<Expr, Expr>`.

Distinct typing of the two expressions has at least two additional
benefits in my eye:

1. It makes the public API more solid and less prone to unexpected usage
   (similar to the internal benefits)

2. Pattern matching is now trivial. Previously, with a common grammar
   our parser erased information about pattern tokens because patterns
   and variables were held in the same kind of expression. This meant we
   had to do additional work to determine the type of a pattern during
   partial evaluation. This work is simple but unnecessary, and by
   propogating pattern tokens into the grammar, matching between
   expression patterns and expressions is now just 3 lines of code.

From this point, I think we can just continue adding more rules, add
support for function rules, and start thinking about beefing up our
parser and handling assignment contexts.

Part of #7


libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -165,28 +172,26 @@ mod tests {
...
-                            scanner_options = scanner_options.set_is_var_char(|c| {
-                                c.is_alphabetic() || c == '$' || c == '#' || c == '_'
-                            });
-                        }
-                        match parse(scan(prog, scanner_options)) {
-                            Stmt::Expr(expr) => expr,
+                    let parse_rule = |prog: &str| -> ExprPat {
+                        let (expr, _) = parse_expression_pattern(scan(prog));
+                        expr
+                    };
```
luke-bhan, line 185: do you think theres a better way to write this. Maybe assert and then return. It seems weird but this may be rust

ayazhafiz: Assert what? That there are no errors?

luke-bhan: the double let just looked awkward to me

ayazhafiz: Oh I see. Well `parse_rule` is a closure to parse a pattern. `let (expr, _)` is to destructure the result of parsing the pattern, which returns an expression and any errors.

luke-bhan: oh, I completely missed the closure part. I understand now. Nice job on this pr!

PR #44, opened 2020-05-10
quick add refactor

luke-bhan:
Flipped the dec vector so that we can work with subtraction and so that its consistent with the int vector.

### DIFF
--- a/libslide/src/bignum.rs
+++ b/libslide/src/bignum.rs
@@ -49,7 +49,7 @@ impl Bignum {
         Bignum {
             is_neg,
             int: int.into_iter().rev().collect(),
-            dec,
+            dec: dec.into_iter().rev().collect(),
         }
     }
 }
@@ -65,7 +65,7 @@ impl fmt::Display for Bignum {
         }
         if !self.dec.is_empty() {
             result.push('.');
-            for n in self.dec.iter() {
+            for n in self.dec.iter().rev() {
                 result.push(to_char(*n));
             }
         }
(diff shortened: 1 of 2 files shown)


luke-bhan:
So the subtraction dec vec doesn't need to be flipped since the carry is propogated in the other direction then the addmethod. And I think its much cleaner to keep them consistent.

ayazhafiz:
Wait, how? I thought the carry is propogated the same way (right to left). Anyway, I think it would be clearer if you open a PR for the sub method and add this change there for more context.

luke-bhan:
Ok, Quick question. I have a leading zeros problem. So 10-1 = 09. How should we go about handling this in your opinion?

ayazhafiz:
I think I'm confused. Can you give me a bigger example with two numbers?

luke-bhan:
1000 - 999 = 0001

luke-bhan:
I think truncate is the way to go

ayazhafiz:
Ohh.. yeah thats unfortunate

ayazhafiz:
Just shift to the front and pop off remaining elements

ayazhafiz:
`pop` will not deallocate: https://doc.rust-lang.org/src/alloc/vec.rs.html#1195-1204

ayazhafiz:
Maybe even easier: reverse, pop, reverse

luke-bhan:
No, we don't need to reverse, our vec is stored for 01000 as 00010. So the zeros are at the end. Thats why I think we can use truncate. It aslso doesnt truncate.

ayazhafiz:
okay sounds good

PR #45, opened 2020-05-10
Added sub for bignum

luke-bhan:
Added sub method. I didn't end up reversing the dec vector since it would have ruined our ability to use truncate in removing ending zeros from the decmial. Added utils in bignum for the is_smaller method, we may need this in other files like divide in the future so I moved it now. lmk what you think @ayazhafiz

### DIFF
--- a/libslide/src/bignum.rs
+++ b/libslide/src/bignum.rs
@@ -1,5 +1,9 @@
 use std::fmt;
 mod add;
+mod compare;
+mod negate;
+mod sub;
+mod utils;
 
 // this probably can go in utils but I put it in here for now
 fn to_u8(c: char) -> u8 {
@@ -17,13 +21,20 @@ pub struct Bignum {
     dec: Vec<u8>, // decimal part with MSB first, e.g. dec of 123.456 is 456.
 }
 
+static INPUT_ERR_MSG: &str = "Input is not valid";
+
 impl Bignum {
-    pub fn new(input: String) -> Bignum {
+    pub fn new(input: String) -> Result<Bignum, &'static str> {
         let mut chars = input.chars();
         let mut is_neg: bool = false;
         let mut has_decimal = false;
         let mut int = Vec::new();
         let mut dec = Vec::new();
+
+        if input.is_empty() {
+            return Err(INPUT_ERR_MSG);
+        }
+
         match chars.next() {
             Some('-') => is_neg = true,
             Some('.') => has_decimal = true,
@@ -40,17 +51,17 @@ impl Bignum {
             } else if c == '.' {
                 has_decimal = !has_decimal;
                 if !has_decimal {
-                    panic!("Number cannot have two decimal values");
+                    return Err(INPUT_ERR_MSG);
                 }
             } else {
-                panic!("Cannot parse string as number");
+                return Err(INPUT_ERR_MSG);
             }
         }
-        Bignum {
+        Ok(Bignum {
             is_neg,
             int: int.into_iter().rev().collect(),
             dec,
-        }
+        })
     }
 }
 
@@ -60,8 +71,12 @@ impl fmt::Display for Bignum {
         if self.is_neg {
             result.push('-');
         }
-        for n in self.int.iter().rev() {
-            result.push(to_char(*n));
+        if self.int.is_empty() {
+            result.push('0');
+        } else {
+            for n in self.int.iter().rev() {
+                result.push(to_char(*n));
+            }
         }
         if !self.dec.is_empty() {
             result.push('.');
@@ -83,7 +98,7 @@ mod tests {
                 use crate::bignum::Bignum;
                 let result = $program.to_string();
                 let bg = Bignum::new(result.clone());
-                assert_eq!(bg.to_string(), result);
+                assert_eq!(bg.unwrap().to_string(), result);
             }
         )*
         }
@@ -96,8 +111,8 @@ mod tests {
             negative_float: "-5.5"
             bigger_int: "1002039444884993020"
             bigger_float: "102399959939999.393993"
-            decimal: ".3"
-            negative_decimal: "-.3"
+            decimal: "0.3"
+            negative_decimal: "-0.3"
         }
     }
 }
--- a/libslide/src/bignum/add.rs
+++ b/libslide/src/bignum/add.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::suspicious_arithmetic_impl)]
+use crate::bignum::utils::abs;
 use crate::bignum::Bignum;
 use std::mem;
 use std::ops;
@@ -7,7 +8,14 @@ impl ops::Add for Bignum {
     type Output = Bignum;
 
     fn add(self, rhs: Bignum) -> Bignum {
-        let is_neg: bool = rhs.is_neg && self.is_neg;
+        let mut is_neg: bool = false;
+        if self.is_neg && rhs.is_neg {
+            is_neg = true;
+        } else if self.is_neg && !rhs.is_neg {
+            return rhs - abs(&self);
+        } else if !self.is_neg && rhs.is_neg {
+            return self - abs(&rhs);
+        }
 
         // this assumes the signs are both positive for now
         let mut carry: u8 = 0;
@@ -23,6 +31,7 @@ impl ops::Add for Bignum {
             (rhs.dec, self.dec)
         };
 
+        // 1. Handle Decimal
         for i in (0..lhs_size).rev() {
             res_dec[i] += lhs_vec[i] + carry;
             carry = (res_dec[i] > 9) as u8;
@@ -81,8 +90,8 @@ mod tests {
             #[test]
             fn $name() {
                 use crate::bignum::Bignum;
-                let lhs = Bignum::new($lhs.to_string());
-                let rhs = Bignum::new($rhs.to_string());
+                let lhs = Bignum::new($lhs.to_string()).unwrap();
+                let rhs = Bignum::new($rhs.to_string()).unwrap();
                 let result = $program.to_string();
                 assert_eq!((lhs+rhs).to_string(), result);
             }
@@ -109,6 +118,18 @@ mod tests {
             mixed: "1.1", "1.1", "2.2"
             mixed2: "1.9999999", "9999.99999", "10001.9999899"
             mixed3: "99999999999999.999999999999", "99999999999.99999", "100099999999999.999989999999"
+            negative_int1: "-5", "5", "0"
+            negative_int2: "-10", "5", "-5"
+            negative_int3: "5", "-10", "-5"
+            negative_int4: "-555555", "999999", "444444"
+            negative_int5: "-10", "-10", "-20"
+            negative_float1: "-0.1", "0.1", "0"
+            negative_float2: "-0.2", "0.1", "-0.1"
+            negative_float3: "-0.1", "0.2", "0.1"
+            negative_float4: "-0.1", "-0.1", "-0.2"
+            negative_mixed: "-12332.55", "1.0", "-12331.55"
+            negative_mixed1: "-12332.55", "-1.0", "-12333.55"
+            negative_mixed2: "1.0", "-12332.55", "-12331.55"
         }
     }
 }
--- a/libslide/src/bignum/compare.rs
+++ b/libslide/src/bignum/compare.rs
@@ -1,0 +1,130 @@
+use crate::bignum::Bignum;
+use std::cmp::Ordering;
+
+impl Ord for Bignum {
+    fn cmp(&self, other: &Bignum) -> Ordering {
+        let greater: Ordering;
+        let less: Ordering;
+        if self.is_neg ^ other.is_neg {
+            return other.is_neg.cmp(&self.is_neg);
+        } else if other.is_neg && self.is_neg {
+            greater = Ordering::Less;
+            less = Ordering::Greater;
+        } else {
+            greater = Ordering::Greater;
+            less = Ordering::Less;
+        }
+        match self.int.len().cmp(&other.int.len()) {
+            Ordering::Greater => return greater,
+            Ordering::Less => return less,
+            _ => (),
+        }
+
+        for i in (0..other.int.len()).rev() {
+            match self.int[i].cmp(&other.int[i]) {
+                Ordering::Greater => return greater,
+                Ordering::Less => return less,
+                _ => (),
+            }
+        }
+
+        let smaller_dec_size: usize = if self.dec.len() > other.dec.len() {
+            other.dec.len()
+        } else {
+            self.dec.len()
+        };
+
+        for i in 0..smaller_dec_size {
+            match self.dec[i].cmp(&other.dec[i]) {
+                Ordering::Greater => return greater,
+                Ordering::Less => return less,
+                _ => (),
+            }
+        }
+
+        match other.dec.len().cmp(&self.dec.len()) {
+            Ordering::Equal => Ordering::Equal,
+            Ordering::Less => less,
+            Ordering::Greater => greater,
+        }
+    }
+}
+
+impl Eq for Bignum {}
+
+impl PartialEq for Bignum {
+    fn eq(&self, other: &Bignum) -> bool {
+        self.cmp(other) == Ordering::Equal
+    }
+}
+
+// returns true if the rhs is smaller then lhs
+impl PartialOrd for Bignum {
+    fn partial_cmp(&self, other: &Bignum) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    macro_rules! bignum_test_cmp {
+        ($($name: ident: $stmt:expr)*) => {
+        $(
+            #[test]
+            fn $name() {
+                use crate::bignum::Bignum;
+                let v: Vec<&str> = $stmt.split(' ').collect();
+                let lhs = Bignum::new(v[0].to_string()).unwrap();
+                let rhs = Bignum::new(v[2].to_string()).unwrap();
+                match v[1] {
+                    "=" => assert!(lhs == rhs),
+                    "!=" => assert!(lhs != rhs),
+                    "<=" => assert!(lhs <= rhs),
+                    ">=" => assert!(lhs >= rhs),
+                    ">" => assert!(lhs > rhs),
+                    "<" => assert!(lhs < rhs),
+                    _ => panic!("Test input invalid"),
+                }
+            }
+        )*
+        }
+    }
+
+    mod cmp {
+        bignum_test_cmp! {
+            equal1: "1 = 1"
+            equal2: "1.1 = 1.1"
+            equal3: "-25.25 = -25.25"
+            not_equal1: "1 != 2"
+            not_equal2: "1.1 != 1.2"
+            not_equal3: "2.1 != 1.1"
+            not_equal4: "-1.1 != 1.1"
+            less_than1: "1 < 2"
+            less_than2: "1.1 < 2.1"
+            less_than3: "2.1 < 2.2"
+            less_than4: "-1 < 1"
+            less_than5: "-2 < -1"
+            less_than_equal1: "1 < 2"
+            less_than_equal2: "1.1 <= 2.1"
+            less_than_equal3: "2.1 <= 2.2"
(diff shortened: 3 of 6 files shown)


libslide/src/bignum.rs
```diff
@@ -24,6 +26,11 @@ impl Bignum {
         let mut has_decimal = false;
         let mut int = Vec::new();
         let mut dec = Vec::new();
+
+        if input.is_empty() {
+            panic!("A arbitray precision number must have a valid input");
+        }
+
```
ayazhafiz, line 33: Maybe it's better to return an option, but we can do that in a different pr

libslide/src/bignum.rs
```diff
@@ -60,8 +67,12 @@ impl fmt::Display for Bignum {
         if self.is_neg {
             result.push('-');
         }
-        for n in self.int.iter().rev() {
-            result.push(to_char(*n));
+        if self.int.is_empty() {
+            result.push('0');
```
ayazhafiz, line 71: Why?

libslide/src/bignum/add.rs
```diff
@@ -23,6 +23,7 @@ impl ops::Add for Bignum {
             (rhs.dec, self.dec)
         };
 
+        // 1. Handle Decimal
```
ayazhafiz, line 26: :+1:

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,148 @@
...
+            if res_int[i] == 0 {
+                count += 1;
+            } else {
+                break;
+            }
+        }
+
+        res_int.truncate(res_int.len() - count);
+
+        print!("{:?}", res_dec);
```
ayazhafiz, line 92: dd

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,148 @@
...
+    type Output = Bignum;
+
+    fn sub(self, other: Bignum) -> Bignum {
+        // make the decimal vectors the same size
+        // fill lhs and rhs vectors as well
+        let mut lhs_dec: Vec<u8>;
+        let mut res_dec: Vec<u8>;
+        let self_size: usize = self.dec.len();
+        let other_size: usize = other.dec.len();
+        let (mut res_int, mut lhs_int) = if self_size > other_size {
```
ayazhafiz, line 18: We can set all four lhs_, res_ vecs here!

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,148 @@
...
+            res_dec = self.dec;
+            lhs_dec = vec![0; self_size];
+            lhs_dec[..other_size].clone_from_slice(&other.dec[..other_size]);
```
ayazhafiz, line 21: I'm confused by this. Why not just
```suggestion
            lhs_dec = other.dec
```

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,148 @@
...
+use std::ops;
+
+impl ops::Sub for Bignum {
+    type Output = Bignum;
+
+    fn sub(self, other: Bignum) -> Bignum {
+        // make the decimal vectors the same size
+        // fill lhs and rhs vectors as well
+        let mut lhs_dec: Vec<u8>;
+        let mut res_dec: Vec<u8>;
```
ayazhafiz, line 15: I think it would be cheaper to set up the vecs as `i8`s here to avoid casting in a loop. LLVM is probably smart enough to optimize it out anyway, but here we make it one step closer and the optimizer can probably vectorize the casts.

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,148 @@
...
+            lhs_dec = vec![0; other_size];
+            lhs_dec[..self_size].clone_from_slice(&self.dec[..self_size]);
+            (other.int, self.int)
+        };
+
+        // now we need to make the bigger number the res dec
+        if is_smaller(&lhs_dec, &res_dec, &lhs_int, &res_int) {
+            mem::swap(&mut lhs_dec, &mut res_dec);
+            mem::swap(&mut lhs_int, &mut res_int);
+        }
```
ayazhafiz, line 34: Why not implement [Ord](https://doc.rust-lang.org/std/cmp/trait.Ord.html) for bignum instead?

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        // 1. Decimal
+
+        let mut res: i8;
+        let mut carry: i8 = 0;
+        for i in (0..res_dec.len()).rev() {
+            res = res_dec[i] as i8 - lhs_dec[i] as i8 - carry;
+            // not sure if you can use a let if here because of shadowing
+            // but maybe you know
```
ayazhafiz, line 43: you can't use a let if, but note that an if can be an expression. That is, you can do something like `a = if true { 1 } else { 2 }`

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,148 @@
...
+            if res < 0 {
+                res += 10;
+                carry = 1;
+            } else {
+                carry = 0;
+            }
```
ayazhafiz, line 49: This is not the prettiest thing in the world, but it does save a branching:

```suggestion
            carry = (res < 0) as i8;
            res += carry * 10; // add 10 if negative, nothing otherwise
```

libslide/src/bignum/sub.rs
ayazhafiz: Got it, nevermind. Is it possible to move a slice into another? If not another option is to set `lhs_dec = other.dec`, then `.reserve()` the rest of the space needed and add 0s. This is pedantic though, the cost savings are just copying some 8-bit values and extra zeros.

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        for i in 0..lhs_int.len() {
+            res = res_int[i] as i8 - lhs_int[i] as i8 - carry;
+            if res < 0 {
+                res += 10;
+                carry = 1;
+            } else {
+                carry = 0;
+            }
+            res_int[i] = res.try_into().unwrap();
```
ayazhafiz, line 63: Yeah let's make res_int a `Vec<i8>`

libslide/src/bignum.rs
luke-bhan: I want decimals to be printed 0.1 instead of .1

libslide/src/bignum/sub.rs
luke-bhan: I will do this instead :)

libslide/src/bignum/sub.rs
luke-bhan: oh sick

libslide/src/bignum/sub.rs
luke-bhan: So technically this wont change the casting. Because either way res_int has to be a u8 in Bignum. so we would have to convert to build it and then convert back to store it

libslide/src/bignum.rs
luke-bhan: how would we retunr an option in the constructor

ayazhafiz: Just change the return type of the method

libslide/src/bignum.rs
ayazhafiz: +1

libslide/src/bignum/sub.rs
ayazhafiz: that's fine, it's cleaner than doing it each time in a loop imo

luke-bhan: so how do you wanna build res_dec. fill with zeros and then cast? I think its equally ugly

luke-bhan: also converting to a vector creates a  new allocation of a vector. I think I like this way better

ayazhafiz: just do it this way: https://stackoverflow.com/questions/59707349/cast-vector-of-i8-to-vector-of-u8-in-rust

luke-bhan:
@ayazhafiz I updated the PR. LMK what you think.

libslide/src/bignum/compare.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+        if self.dec.len() != other.dec.len() {
+            return false;
+        }
+        for i in 0..other.dec.len() {
+            if other.dec[i] != self.dec[i] {
+                return false;
+            }
+        }
+        true
```
ayazhafiz, line 62: Why not just

```rust
self.cmp(other) == Ordering::Equal
```

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,145 @@
...
+        let mut lhs = Bignum {
+            is_neg: false,
+            dec: lhs_dec,
+            int: lhs_int,
+        };
+        let mut rhs = Bignum {
+            is_neg: false,
+            dec: res_dec,
+            int: res_int,
+        };
```
ayazhafiz, line 37: let's not construct a new bignum just to do the swap

libslide/src/bignum/utils.rs
```diff
@@ -0,0 +1,15 @@
...
+    unsafe { Vec::from_raw_parts(p as *mut u8, len, cap) }
+}
+
+pub fn vec_u8_into_i8(v: Vec<u8>) -> Vec<i8> {
+    let mut v = std::mem::ManuallyDrop::new(v);
+    let p = v.as_mut_ptr();
+    let len = v.len();
+    let cap = v.capacity();
+    unsafe { Vec::from_raw_parts(p as *mut i8, len, cap) }
+}
```
ayazhafiz, line 15: We can combine this

```suggestion
pub fn recast_vec<T, U>(v: Vec<T>) -> Vec<U> {
    let mut v = std::mem::ManuallyDrop::new(v);
    let p = v.as_mut_ptr();
    let len = v.len();
    let cap = v.capacity();
    unsafe { Vec::from_raw_parts(p as *mut U, len, cap) }
}
```

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,145 @@
...
+        // Propogate the rest of the carry
+        let mut i = lhs_int_asi8.len();
+        while i < res_int_asi8.len() {
+            res = res_int_asi8[i] as i8 - carry;
+            if res < 0 {
+                res += 10;
+                carry = 1;
+            } else {
+                carry = 0;
+            }
```
ayazhafiz, line 74: Suggestion from before

```suggestion
carry = res < 0;
res += 10 * carry;
```

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,145 @@
...
+            if res_dec_asi8[i] == 0 {
+                count += 1;
+            } else {
+                break;
+            }
+        }
+        res_dec_asi8.truncate(res_dec_asi8.len() - count);
+
+        Bignum {
+            is_neg: false,
```
ayazhafiz, line 102: Do you plan to handle negatives in another PR?

ayazhafiz:
@luke-bhan left some comments, mostly minor stuff

libslide/src/bignum/sub.rs
luke-bhan: Yes

libslide/src/bignum/sub.rs
luke-bhan: It’s not to do the swap, it’s to compare. What’s a better way of doing this then.

ayazhafiz: nevermind, sounds good 👍

ayazhafiz: Actually, hold up. Both self and rhs are moved in at the beginning of this function, so why not just compare them then before you construct the vectors?

libslide/src/bignum/compare.rs
luke-bhan: What do you mean

libslide/src/bignum/sub.rs
luke-bhan: That’s how I originally wrote sub, but Then you run into an issue with adding the zeros to make the decimals the same length.

ayazhafiz: What's wrong with that?

libslide/src/bignum/compare.rs
ayazhafiz: Replace this whole method with `self.cmp(other) == Ordering::Equal`

luke-bhan:
ok, @ayazhafiz everything should be fixed except tests for ordering. I will add that and negatives in a quick fixup pr in a few hours (I don't know how long its gonna take hopefully less)

luke-bhan:
Ok, I actually added negatives and compares in this pr. (I'm lazy and don't want to make an extra one lol)

libslide/src/bignum/add.rs
```diff
@@ -7,7 +8,14 @@ impl ops::Add for Bignum {
     type Output = Bignum;
 
     fn add(self, rhs: Bignum) -> Bignum {
-        let is_neg: bool = rhs.is_neg && self.is_neg;
+        let mut is_neg: bool = false;
+        if self.is_neg && rhs.is_neg {
+            is_neg = true;
+        } else if self.is_neg && !rhs.is_neg {
+            return rhs - abs(&self);
+        } else if !self.is_neg && rhs.is_neg {
```
ayazhafiz, line 16: You only need to do one comparison on each branch
```suggestion
        } else if self.is_neg {
            return rhs - abs(&self);
        } else if rhs.is_neg {
```

libslide/src/bignum/add.rs
```diff
@@ -7,7 +8,14 @@ impl ops::Add for Bignum {
     type Output = Bignum;
 
     fn add(self, rhs: Bignum) -> Bignum {
-        let is_neg: bool = rhs.is_neg && self.is_neg;
+        let mut is_neg: bool = false;
+        if self.is_neg && rhs.is_neg {
+            is_neg = true;
```
ayazhafiz, line 13:
```suggestion
        let mut is_neg: bool = self.is_neg && rhs.is_neg;
        if self.is_neg && rhs.is_neg {
            is_neg = true;
```

And just get rid of this branch

libslide/src/bignum/compare.rs
```diff
@@ -0,0 +1,131 @@
...
+            return Ordering::Less;
+        } else if other.is_neg && !self.is_neg {
+            return Ordering::Greater;
+        } else if other.is_neg && self.is_neg {
+            greater = Ordering::Less;
+            less = Ordering::Greater;
+        } else {
+            greater = Ordering::Greater;
+            less = Ordering::Less;
+        }
```
ayazhafiz, line 18: Why include these branches?

libslide/src/bignum/compare.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::cmp::Ordering;
+
+impl Ord for Bignum {
+    fn cmp(&self, other: &Bignum) -> Ordering {
+        let greater: Ordering;
+        let less: Ordering;
+        if self.is_neg && !other.is_neg {
+            return Ordering::Less;
+        } else if other.is_neg && !self.is_neg {
+            return Ordering::Greater;
```
ayazhafiz, line 11: Do you want to see a way to save a branch in a somewhat convoluted manner? Great, you're in luck!
```suggestion
        if self.is_neg ^ other.is_neg {
             return other.is_neg.cmp(&self.is_neg)
        }
```

libslide/src/bignum/compare.rs
```diff
@@ -0,0 +1,131 @@
+use crate::bignum::Bignum;
+use std::cmp::Ordering;
+
+impl Ord for Bignum {
+    fn cmp(&self, other: &Bignum) -> Ordering {
+        let greater: Ordering;
+        let less: Ordering;
```
ayazhafiz, line 7: nit: prefer using `Ordering::` directly rather than variables

libslide/src/bignum/compare.rs
```diff
@@ -0,0 +1,131 @@
...
+                    _ => panic!("Test input invalid"),
+                }
+            }
+        )*
+        }
+    }
+
+    mod cmp {
+        bignum_test_cmp! {
+            equal1: "1", "1", "="
```
ayazhafiz, line 96: Beautiful! Just one suggestion - write the test as `1 = 1` and split by spaces into three components to compare with.

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        // now we need to make the bigger number the res dec
+        // 1. Decimal
+
+        let mut res: i8;
+        let mut carry: i8 = 0;
+        let mut res_dec_asi8 = recast_vec::<u8, i8>(res_dec);
+        let lhs_dec_asi8 = recast_vec::<u8, i8>(lhs_dec);
+        let mut res_int_asi8 = recast_vec::<u8, i8>(res_int);
+        let lhs_int_asi8 = recast_vec::<u8, i8>(lhs_int);
```
ayazhafiz, line 57: I think it would make more sense to do this in lines 39-46.

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,170 @@
...
+            if self.is_neg {
+                is_neg = true;
+            }
+            lhs = other;
+            rhs = self;
+        } else {
+            // this really is !self.is_neg && !other.is_neg but we can use demorgans for speed
+            if !(self.is_neg || other.is_neg) {
+                is_neg = true;
+            }
```
ayazhafiz, line 30: It would be even faster if you set this on line 11 and avoided the branch.

libslide/src/bignum/sub.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        // quick optimization to remove -0 from possible results.
+        if is_neg && res_int_asi8.is_empty() && res_dec_asi8.is_empty() {
+            is_neg = false;
+        }
+
+        Bignum {
+            is_neg,
+            dec: recast_vec::<i8, u8>(res_dec_asi8),
+            int: recast_vec::<i8, u8>(res_int_asi8),
```
ayazhafiz, line 114: I think you can elide the type parameters

libslide/src/bignum/compare.rs
luke-bhan: because if its a negative number we need to switch the comparison. because we are comparing the magnitude in all the branches? I know this isn't a great way to handle it but I couldn't really think of anything cleaner. Do you have any general ideas?

libslide/src/bignum/sub.rs
luke-bhan: how? doesn't it need to know what type to cast to?

libslide/src/bignum/sub.rs
luke-bhan: yep. will do

libslide/src/bignum/sub.rs
luke-bhan: you can't do this, I am already avoiding branches by setting it to false on line 11

libslide/src/bignum/compare.rs
ayazhafiz: nvm this is fine

libslide/src/bignum/sub.rs
ayazhafiz: Try it

libslide/src/bignum/sub.rs
ayazhafiz: Why can't you do it? Just do is_neg = self.is_neg || other.is_neg.

luke-bhan: oh i see ok.

luke-bhan: I was thinking  a different idea

luke-bhan: this idea doesn't work. It needs to be a specific case where the lhs is < rhs to work. I think the way I have it is fine.

ayazhafiz: How can that be the case? If they're both negative numbers the result must be negative.

luke-bhan: no, this is handling if  they are both positive and lhs is < rhs.

ayazhafiz: got it

PR #47, opened 2020-05-11
Add support for function rules

### LINKED ISSUE #43: Support function rules
Pattern-mapping rules like we currently have aren't enough to simplify constant expressions like `1 + 2`, which need an implementation for addition. We should add support for this.

ayazhafiz:
Also adds some additional pattern reordering rules.

There are a lot of todos for us to continue working on -- see comments.
I'll add issues for them after the PR for this commit is merged.

Closes #43


libslide/src/evaluator_rules/registry.rs
```diff
@@ -0,0 +1,90 @@
...
+    UR::F(divide),                        // Divide
+    UR::F(modulo),                        // Modulo
+    UR::F(exponentiate),                  // Exponentiate
+    UR::F(posate),                        // Posate
+    UR::F(negate),                        // Negate
+    UR::S("_a + 0 -> _a"),                // AdditiveIdentity
+    UR::M(&["#a + $b -> $b + #a"]),       // ReorderConstants
+    UR::M(&["($a) -> $a", "(#a) -> #a"]), // UnwrapParens
+    UR::M(&["[$a] -> $a", "[#a] -> #a"]), // UnwrapBraces
+];
```
luke-bhan, line 38: UR is not a great name. Lets use unbuilt_rule or something.

libslide/src/evaluator_rules/unbuilt_rule.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// rather than the string representations of expressions. This ensures rule application is
+    /// always exact and deterministic.
+    S(&'static str),
+
+    /// Multiple expression mappings representing one rule, perhaps in several different ways.
+    M(&'static [&'static str]),
+
+    /// A function rule.
+    F(fn(Expr) -> Option<Expr>),
+}
```
luke-bhan, line 45: These letters are confusing, I think we should expand on their naming. I didn't know what m at all intuitively until I read the comment.

luke-bhan: Can you also give an example of S v.s. M, I am not sure I understand.

libslide/src/evaluator_rules/registry.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    #[test]
+    fn builds_rules() {
+        let rule_set = RuleSet::default();
+        let built_rules = rule_set.build();
+        let var_plus_zero = &built_rules[RuleName::AdditiveIdentity as usize];
+
+        assert_eq!(var_plus_zero.to_string(), "_a + 0 -> _a");
+    }
+}
```
luke-bhan, line 90: this should be cleaned up somehow for when we have a large number of rules. I assume you plan on this.

luke-bhan: Can we also add tseting for multiple rules.

libslide/src/partial_evaluator.rs
```diff
@@ -56,12 +56,56 @@ mod tests {
...
+
+        exp:                            "2 ^ 3"     => "8"
+        exp_nested_left:                "2 ^ 3 ^ a" => "2 ^ 3 ^ a"
+        exp_associated:                 "2 ^ 3 ^ 2" => "512"
+
+        posate:                         "+1"           => "1"
+        // TODO: unwrap parantheses further
+        posate_nested_1:                "+(b + c)" => "(b + c)"
+        posate_nested:                  "a + +(b + c)" => "a + (b + c)"
+
```
luke-bhan, line 88: add negative versions of these. The positive version is relatively imo and I still don't think it was necessary to add a unary operator for it. I think it is bad to allow our users to have a functiion  ++++5 without throwing an error saying they are an idiot.

libslide/src/evaluator_rules/registry.rs
ayazhafiz: Would you be comfortable with changing to `UnbuiltRule` but importing it as `UR`? I just wanted to do it for succinctness so we could fit all this in the array.

libslide/src/evaluator_rules/unbuilt_rule.rs
ayazhafiz: Yeah that's a good point. I was hoping the comments would be enough.

`M` is for when we need multiple patterns to express one rule. For example we can unwrap parentheses for either just constants or just variables, but we have no pattern that describes constant|variable, so we have to use two patterns -- `"($a) -> $a"` and `"(#a) -> #a"`.

libslide/src/evaluator_rules/registry.rs
ayazhafiz: Well this function just checks a single rule is built. Would you like to check that we build all rules correctly? It should be easy to do that.

libslide/src/partial_evaluator.rs
ayazhafiz: Can you elaborate by what you mean by negative versions? In particular, what rules would you like? I can think of `"-(_a - _b) -> _b - _a"`, anything else?

luke-bhan: -(b + c) => -b -c is all I was thinking. This can be added in another pr if you want

libslide/src/evaluator_rules/registry.rs
luke-bhan: I think your update is fine

libslide/src/evaluator_rules/unbuilt_rule.rs
luke-bhan: gotcha.

libslide/src/evaluator_rules/registry.rs
luke-bhan: yea thats all.

luke-bhan: Gonna approve, just add a quick test here for building all rules.

libslide/src/partial_evaluator.rs
ayazhafiz: I'll add it

PR #48, opened 2020-05-11
Common subexpression elimination

### LINKED ISSUE #46: Common subexpression elimination
https://en.wikipedia.org/wiki/Common_subexpression_elimination

CSE should be trivial to implement for us.

@luke-bhan do you want to do this?

ayazhafiz:
We do this by passing reference-counted pointers around in place of
expression objects.

I'm actually not sure if this is worth the overhead compared to what we
already had. @luke-bhan could you give me a thorough review on this one?

Closes #46

### DIFF
--- a/libslide/src/grammar/pattern.rs
+++ b/libslide/src/grammar/pattern.rs
@@ -1,5 +1,7 @@
 pub use super::*;
 
+use std::rc::Rc;
+
 #[derive(Clone)]
 pub enum ExprPat {
     Const(f64),
@@ -11,11 +13,12 @@ pub enum ExprPat {
     AnyPat(String),
     BinaryExpr(BinaryExpr<Self>),
     UnaryExpr(UnaryExpr<Self>),
-    Parend(Box<Self>),
-    Braced(Box<Self>),
+    Parend(Rc<Self>),
+    Braced(Rc<Self>),
 }
 
 impl Grammar for ExprPat {}
+impl Grammar for Rc<ExprPat> {}
 impl Expression for ExprPat {}
 
 // TODO: We can't derive this because `f64` doesn't implement `Eq`.
@@ -42,7 +45,19 @@ impl PartialEq for ExprPat {
 // arbitrary-precision numeric type.
 impl core::hash::Hash for ExprPat {
     fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
-        state.write(self.to_string().as_bytes())
+        use ExprPat::*;
+        match self {
+            // TODO: We can do better than hashing to a string as well, but we'll save that til we
+            // have an arbitrary-precision numeric type.
+            Const(f) => state.write(f.to_string().as_bytes()),
+            VarPat(v) => v.hash(state),
+            ConstPat(c) => c.hash(state),
+            AnyPat(a) => a.hash(state),
+            BinaryExpr(e) => e.hash(state),
+            UnaryExpr(e) => e.hash(state),
+            e @ Parend(_) => e.to_string().hash(state),
+            e @ Braced(_) => e.to_string().hash(state),
+        }
     }
 }
 
--- a/libslide/src/parser.rs
+++ b/libslide/src/parser.rs
@@ -13,11 +13,12 @@ use crate::utils::PeekIter;
 
 use core::convert::TryFrom;
 use core::fmt::Display;
+use std::rc::Rc;
 
 macro_rules! binary_expr_parser {
     ($self:ident $($name:ident: lhs=$lhs_term:ident, rhs=$rhs_term:ident, op=[$($matching_op:tt)+])*) => {
         $(
-        fn $name(&mut $self) -> Box<Self::Expr> {
+        fn $name(&mut $self) -> Rc<Self::Expr> {
             use BinaryOperator::*;
 
             let mut lhs = $self.$lhs_term();
@@ -29,11 +30,12 @@ macro_rules! binary_expr_parser {
                 match op {
                     $($matching_op)+ => {
                         $self.input().next();
-                        lhs = Box::new(BinaryExpr{
+                        let bin_expr = BinaryExpr{
                             op,
                             lhs,
                             rhs: $self.$rhs_term(),
-                        }.into());
+                        }.into();
+                        lhs = $self.finish_expr(bin_expr);
                     }
                     _ => break,
                 }
@@ -47,16 +49,16 @@ macro_rules! binary_expr_parser {
 trait Parser<T>
 where
     T: Grammar,
-    Self::Expr: Expression,
+    Self::Expr: Expression + Clone,
     Self::Error: Display,
 {
     type Expr;
     type Error;
 
     fn new(input: Vec<Token>) -> Self;
-    fn parse(&mut self) -> Box<T>;
     fn errors(&self) -> &Vec<Self::Error>;
     fn input(&mut self) -> &mut PeekIter<Token>;
+    fn parse(&mut self) -> T;
     fn parse_float(&mut self, f: f64) -> Self::Expr;
     fn parse_variable(&mut self, name: String) -> Self::Expr;
     fn parse_var_pattern(&mut self, name: String) -> Self::Expr;
@@ -64,6 +66,7 @@ where
     fn parse_any_pattern(&mut self, name: String) -> Self::Expr;
     fn parse_open_paren(&mut self) -> Self::Expr;
     fn parse_open_brace(&mut self) -> Self::Expr;
+    fn finish_expr(&mut self, expr: Self::Expr) -> Rc<Self::Expr>;
 
     // Default parsing implementations.
     // TODO: increase modularity of this parser
@@ -72,7 +75,7 @@ where
         self.input().peek().map(|t| &t.ty) == Some(&TokenType::EOF)
     }
 
-    fn expr(&mut self) -> Box<Self::Expr> {
+    fn expr(&mut self) -> Rc<Self::Expr> {
         self.add_sub_term()
     }
 
@@ -89,35 +92,34 @@ where
         exp_term:            lhs = num_term,            rhs = exp_term,            op = [Exp]
     );
 
-    fn num_term(&mut self) -> Box<Self::Expr> {
+    fn num_term(&mut self) -> Rc<Self::Expr> {
         if let Some(Ok(op)) = self.input().peek().map(UnaryOperator::try_from) {
             self.input().next();
-            return Box::new(
-                UnaryExpr {
-                    op,
-                    rhs: self.exp_term(),
-                }
-                .into(),
-            );
+            let node = UnaryExpr {
+                op,
+                rhs: self.exp_term(),
+            }
+            .into();
+            return self.finish_expr(node);
         }
 
         let node = match self.input().peek().map(|t| t.ty.clone()) {
-            Some(TokenType::Float(f)) => self.parse_float(f).into(),
-            Some(TokenType::Variable(name)) => self.parse_variable(name).into(),
+            Some(TokenType::Float(f)) => self.parse_float(f),
+            Some(TokenType::Variable(name)) => self.parse_variable(name),
 
             // TODO: Currently patterns as all parsed as variables. Maybe there is a nice way to
             // store pattern information during parsing, but this would require some extension of
             // the grammar.
-            Some(TokenType::VariablePattern(name)) => self.parse_var_pattern(name).into(),
-            Some(TokenType::ConstPattern(name)) => self.parse_const_pattern(name).into(),
-            Some(TokenType::AnyPattern(name)) => self.parse_any_pattern(name).into(),
+            Some(TokenType::VariablePattern(name)) => self.parse_var_pattern(name),
+            Some(TokenType::ConstPattern(name)) => self.parse_const_pattern(name),
+            Some(TokenType::AnyPattern(name)) => self.parse_any_pattern(name),
 
-            Some(TokenType::OpenParen) => self.parse_open_paren().into(),
-            Some(TokenType::OpenBracket) => self.parse_open_brace().into(),
+            Some(TokenType::OpenParen) => self.parse_open_paren(),
+            Some(TokenType::OpenBracket) => self.parse_open_brace(),
             _ => unreachable!(),
         };
         self.input().next(); // eat rest of created expression
-        node
+        self.finish_expr(node)
     }
 }
 
@@ -159,6 +161,7 @@ mod tests {
         parentheses_exp_time:    "2 ^ (3 ^ 4 * 5)"
         parentheses_unary:       "-(2 + +-5)"
         nested_parentheses:      "((1 * (2 + 3)) ^ 4)"
+        bracket1: "[1 + 2]"
         brackets_plus_times:     "[1 + 2] * 3"
         brackets_time_plus:      "3 * [1 + 2]"
         brackets_time_mod:       "3 * [2 % 2]"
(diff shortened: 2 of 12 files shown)


luke-bhan:
jesus this is a big one. Looking through it now.

luke-bhan:
Generally, I think this will be useful bc arbitrary addition, subtraction and multiplication isn't going to be super fast compared to the storage space.

libslide/src/evaluator_rules.rs
```diff
@@ -109,36 +112,64 @@ impl Transformer<Expr, Expr> for BuiltRule {
...
-                rhs: self.transform(*binary_expr.rhs).into(),
-            }
-            .into(),
-            Expr::UnaryExpr(unary_expr) => UnaryExpr {
-                op: unary_expr.op,
-                rhs: self.transform(*unary_expr.rhs).into(),
+    fn transform(&self, target: Rc<Expr>) -> Rc<Expr> {
+        fn fill(cache: &mut HashMap<u64, Rc<Expr>>, t: Rc<Expr>, r: Rc<Expr>) -> Rc<Expr> {
+            Rc::clone(cache.entry(hash(t.as_ref())).or_insert_with(|| r))
+        }
```
luke-bhan, line 118: why do we need to use Rc::clone. Can't we just originally create an Rc pointer instead of allocating a new one?

libslide/src/evaluator_rules.rs
```diff
@@ -109,36 +112,64 @@ impl Transformer<Expr, Expr> for BuiltRule {
...
+
+        fn transform(
+            rule: &BuiltRule,
+            target: Rc<Expr>,
+            cache: &mut HashMap<u64 /* ptr address */, Rc<Expr>>,
+        ) -> Rc<Expr> {
```
luke-bhan, line 124: move the comment here, it makes the code less readable.

ayazhafiz:
Cool cool

libslide/src/evaluator_rules.rs
```diff
@@ -109,36 +112,64 @@ impl Transformer<Expr, Expr> for BuiltRule {
...
+                // Could not match the rule on the top-level of the expression; return the partially
+                // transformed expression.
+                None => return fill(cache, target, partially_transformed),
+            };
+
+            // If the rule was matched on the expression, we have replacements for rule patterns ->
+            // target subexpressions. Apply the rule by transforming the rule's RHS with the
+            // replacements.
+            let transformed = replacements.transform(rule.to.clone().into());
+            fill(cache, target, transformed)
```
luke-bhan, line 168: I think this match statement can be conslidated with the match expression above. so that we just return a tranformed repl.

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -59,47 +64,72 @@ pub fn match_rule(rule: ExprPat, target: Expr) -> Option<Replacements> {
...
-                    // A pattern can never be transformed to an expression if it is not replaced!
-                    // TODO: Return an error rather than panicking
-                    None => unreachable!(),
-                }
+    fn transform(&self, item: Rc<ExprPat>) -> Rc<Expr> {
+        fn transform(
+            repls: &Replacements,
+            item: Rc<ExprPat>,
+            cache: &mut HashMap<u64 /* ptr address */, Rc<Expr>>,
+        ) -> Rc<Expr> {
```
luke-bhan, line 85: again this comment is awkward. You already said u64 is the address above in line 67

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -59,47 +64,72 @@ pub fn match_rule(rule: ExprPat, target: Expr) -> Option<Replacements> {
...
+                cache
+                    .entry(hash(item.as_ref()))
+                    .or_insert_with(|| transformed),
+            )
         }
+
+        let mut cache = HashMap::new();
+        transform(self, item, &mut cache)
     }
 }
```
luke-bhan, line 135: this is a very clean function

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -122,78 +152,109 @@ impl Replacements {
         Some(replacements)
     }
 
-    fn insert(&mut self, k: ExprPat, v: Expr) -> Option<Expr> {
-        self.map.insert(k, v)
+    fn insert(&mut self, k: &Rc<ExprPat>, v: Rc<Expr>) -> Option<Rc<Expr>> {
+        self.map.insert(hash(k.as_ref()), v)
     }
```
luke-bhan, line 157: isn't k already a reference from its definition in line 155. why do we need it again as_ref()

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -122,78 +152,109 @@ impl Replacements {
...
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::{parse_expression, parse_expression_pattern, scan};
+
+    fn parse_rule(prog: &str) -> ExprPat {
+        let (expr, _) = parse_expression_pattern(scan(prog));
+        expr
+    }
+
```
luke-bhan, line 169: can't we just return parse_expression_pattern.unwrap()?

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -122,78 +152,109 @@ impl Replacements {
...
+    fn parse_expr(prog: &str) -> Expr {
+        match parse_expression(scan(prog)) {
+            (Stmt::Expr(expr), _) => expr,
+            _ => unreachable!(),
+        }
```
luke-bhan, line 174: I don't understand the meaning for this statement. Shouldn't you use if let since all you're doing is returning the enum variant?

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -122,78 +152,109 @@ impl Replacements {
...
             let merged = Replacements::try_merge(left, right);
             assert!(merged.is_none());
         }
+
+        #[test]
+        fn transform_common_subexpression_elimination() {
+            let parsed_rule_from = parse_rule("#a * _b + #a * _b");
+            let parsed_rule_to = parse_rule("#a * _b + #a * _b");
+            let parsed_target = parse_expr("0 * 0 + 0 * 0");
+
```
luke-bhan, line 220: why are parsed_rule_from and parsed_rule_to the same thing. Just set one variable then. If this is just for the exercise, add a comment for the future.

libslide/src/evaluator_rules/pattern_matching.rs
```diff
@@ -122,78 +152,109 @@ impl Replacements {
...
+            let repls = match_rule(Rc::new(parsed_rule_from), Rc::new(parsed_target)).unwrap();
+            let transformed = repls.transform(Rc::new(parsed_rule_to));
+            let (l, r) = match transformed.as_ref() {
+                Expr::BinaryExpr(BinaryExpr { lhs, rhs, .. }) => (lhs, rhs),
+                _ => unreachable!(),
+            };
+            assert!(std::ptr::eq(l.as_ref(), r.as_ref())); // #a * _b
+
```
luke-bhan, line 228: again in 223-225 youre just pulling the enum variant out. Why use a match?

libslide/src/grammar/pattern.rs
```diff
@@ -11,11 +13,12 @@ pub enum ExprPat {
     AnyPat(String),
     BinaryExpr(BinaryExpr<Self>),
     UnaryExpr(UnaryExpr<Self>),
-    Parend(Box<Self>),
-    Braced(Box<Self>),
+    Parend(Rc<Self>),
+    Braced(Rc<Self>),
 }
```
luke-bhan, line 18: I know this is in a previous PR, but why are these called parend and braced instead of paren and bracket.

libslide/src/parser.rs
```diff
@@ -89,35 +91,34 @@ where
...
-                }
-                .into(),
-            );
+            let node = UnaryExpr {
+                op,
+                rhs: self.exp_term(),
+            }
+            .into();
+            return self.finish_expr(node);
         }
```
luke-bhan, line 103: is this doable without creating a temp var

libslide/src/parser/expression_pattern_parser.rs
```diff
@@ -101,4 +119,30 @@ mod tests {
...
+                ExprPat::BinaryExpr(BinaryExpr {
+                    lhs: rl, rhs: rr, ..
+                }),
+            ) => (ll, lr, rl, rr),
+            _ => unreachable!(),
+        };
+        assert!(std::ptr::eq(ll.as_ref(), rl.as_ref())); // 1
+        assert!(std::ptr::eq(lr.as_ref(), rr.as_ref())); // 2
+    }
 }
```
luke-bhan, line 148: This may be just an idea, but can't we declare a few macros here for finish_expr to avoid writing the same function twice in expression_pattern_parser, and expression_parser?

libslide/src/partial_evaluator.rs
```diff
@@ -1,39 +1,33 @@
...
 pub fn evaluate(expr: Stmt) -> Expr {
     let rule_set = RuleSet::default();
     let built_rules = rule_set.build();
 
-    let mut simplified_expr: Expr = match expr {
-        Stmt::Expr(expr) => expr,
+    let mut simplified_expr: Rc<Expr> = match expr {
+        Stmt::Expr(expr) => expr.into(),
         _ => todo!("Evaluation currently only handles expressions"),
     };
```
luke-bhan, line 18: add a // todo comment here so we are consistent with our other todos

libslide/src/utils/hash.rs
```diff
@@ -0,0 +1,9 @@
+use core::hash::{Hash, Hasher};
+use std::collections::hash_map::DefaultHasher;
+
+pub fn hash<T: Hash>(expr: &T) -> u64 {
+    // There is no way to reset a hasher's state, so we create a new one each time.
+    let mut hasher = DefaultHasher::new();
+    expr.hash(&mut hasher);
+    hasher.finish()
+}
```
luke-bhan, line 9: This is better in utils nice!

luke-bhan:
Ok, I went through line by line :). Some of my comments may be a misunderstanding on my part or flat out wrong, but hopefully theres some valuable stuff in there @ayazhafiz

libslide/src/evaluator_rules.rs
ayazhafiz: This is the great thing about `Rc`! Cloning gives a new reference to the inner pointer, not actually cloning the inner data.

libslide/src/evaluator_rules.rs
ayazhafiz: Move it where?

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: Yeah I'll remove this

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: It's a reference to the `Rc`, not the inner data. We call `as_ref` to be a reference (pointer) to the underlying data, which is the common subexpression.

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: No, `parse_expression_pattern` returns a tuple which we have to destruct here.

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: No, we're returning data inner to the enum variant which is inner to the tuple returned by `parse_Expression`. Even if we use `if let`, we have to handle the default (unreachable case)

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: Good spot, I'll make them the same var

libslide/src/evaluator_rules/pattern_matching.rs
ayazhafiz: See a couple of comments above.

1. To pull an enum variant out we need to perform pattern matching anyway (there is no other way save memory discrimination, which is << pattern matching IMO)

2. We are then pulling out a `BinaryExpr` held by the variant, which also requires matching.

libslide/src/grammar/pattern.rs
ayazhafiz: Because they are paranthesized and braced expressions, not a parantheses and bracket.

libslide/src/parser.rs
ayazhafiz: Yes but it's a bunch of nested calls that spans >5 lines, which I would prefer to avoid. This will definitely get optimized out anyway

libslide/src/parser/expression_pattern_parser.rs
ayazhafiz: I would actually prefer not doing that because the macro would have to live in a file different from both of these, which I don't think is worth doing for just one macro and such a small function. Let me know what you think.

libslide/src/partial_evaluator.rs
ayazhafiz: will do

libslide/src/evaluator_rules.rs
ayazhafiz: How? First we need to apply the rule recursively (first match expression), then rule pattern matching.

libslide/src/evaluator_rules.rs
luke-bhan: just get rid of it.

libslide/src/evaluator_rules.rs
luke-bhan:
```
return match match_rule (rule.from.clone().into(), (*partially_transformed).clone().into() {
Some(repl) => let tranformed = repl.tranform(rule.to.clone().into()); fill(cache, target, transformed)
None => fill(cache, target, partially_transformed)
}
```

luke-bhan: something like that. what do you think

libslide/src/grammar/pattern.rs
luke-bhan: I still don't like this naming. Is this what other libraries use? Braced does not intuitively make me think of a bracket expression.

libslide/src/parser/expression_pattern_parser.rs
luke-bhan: thats fine, these will probably be edited anyway

libslide/src/grammar/pattern.rs
ayazhafiz: Yeah Braced needs to be changed, can we do it in another pr?

libslide/src/evaluator_rules.rs
ayazhafiz: i see now. done, thanks

libslide/src/evaluator_rules.rs
ayazhafiz: I think it's a valuable comment to explain why we're indexing against an integer though

luke-bhan: ok, I just think its inconsitent that we have comments in places where things are used like here and in where we define the Hashmap whereas when dealing with the definitions of unbuilt rules, there's only comments in the implementation of UR and not when UR is being used.

libslide/src/grammar/pattern.rs
luke-bhan: yes

libslide/src/evaluator_rules.rs
ayazhafiz: oh true, removed, forgot that the cache is stored below

PR #52, opened 2020-05-13
Complex Numbers for floats.

luke-bhan:
Added support for complex numbers in the form of floats. We will need this for the fft. Yes, I know this is a lot of work for one op but it transforms the operation from n^2 to nlogn (theoretically, I've read its not as perfect as expected). Also, the complex file can be abstracted out, but I wasn't sure it was necessary. LMK what you think. @ayazhafiz 

Also on another note, we should probably support complex numbers in our expression tree at some point!

### DIFF
--- a/libslide/src/bignum/complex.rs
+++ b/libslide/src/bignum/complex.rs
@@ -1,0 +1,222 @@
+#![allow(clippy::suspicious_arithmetic_impl)]
+#![allow(dead_code)]
+use std::convert::TryFrom;
+use std::fmt;
+use std::ops;
+
+static INPUT_ERR_MSG: &str = "Input is not valid";
+const TOLERANCE: f64 = 1E-12;
+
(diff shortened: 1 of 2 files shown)


libslide/src/bignum/complex.rs
```diff
@@ -0,0 +1,218 @@
...
+use std::fmt;
+use std::ops;
+
+static INPUT_ERR_MSG: &str = "Input is not valid";
+const TOLERANCE: f64 = 1E-12;
+
+#[derive(Clone, Copy)]
+struct Complex {
+    real: f64,
+    imag: f64,
```
ayazhafiz, line 13: Could we use Bignums here in the future?

luke-bhan: No, because then bignums multiplication would be circular. I intend to make a complex class for bignums in the future.

PR #53, opened 2020-05-13
added support for fft

luke-bhan:
I feel our pull requests have been big so heres a small one. FFT added to utils.rs

### DIFF
--- a/libslide/src/bignum/utils.rs
+++ b/libslide/src/bignum/utils.rs
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
+use crate::bignum::complex::Complex;
 use crate::bignum::Bignum;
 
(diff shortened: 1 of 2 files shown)


libslide/src/bignum/utils.rs
```diff
@@ -15,3 +17,35 @@ pub fn abs(rhs: &Bignum) -> Bignum {
         int: rhs.int.clone(),
     }
 }
+
+pub fn fft(item: Vec<Complex>, n: usize) -> Vec<Complex> {
```
ayazhafiz, line 21: tests?

luke-bhan: Shit forgot. Will add

luke-bhan:
I just added a super simple test, but I really don't think we need more. PTAL @ayazhafiz

PR #57, opened 2020-05-14
Generalize parantheses unwrapping and bootstrap rules

### LINKED ISSUE #54: Unwrap parentheses in low-precedence contexts
Our parantheses-unwrapping logic is somewhat naive. We only unwrap expressions that are exactly constants or variables. This means we lose unwrapping other expressions in contexts where they could be. For example, the following two expressions are not simplified further, though they should be:

```
(b + c)     // should => b + c
a + (b + c) // should => a + b + c
a + (b * c) // should => a + b * c
```

ayazhafiz:
Closes #54

### DIFF
--- a/libslide/src/evaluator_rules/registry.rs
+++ b/libslide/src/evaluator_rules/registry.rs
@@ -1,6 +1,6 @@
 mod fn_rules;
 
-use super::rule::Rule;
+use super::rule::*;
 use super::unbuilt_rule::UnbuiltRule;
 use fn_rules::*;
 
@@ -33,8 +33,8 @@ define_rules! {
                AdditiveIdentity: S("_a + 0 -> _a")
                ReorderConstants: S("#a + $b -> $b + #a")
              DistributeNegation: S("-(_a - _b) -> _b - _a")
-                   UnwrapParens: M(&["($a) -> $a", "(#a) -> #a"])
-                   UnwrapBraces: M(&["[$a] -> $a", "[#a] -> #a"])
+           UnwrapExplicitParens: S("(_a) -> _a")
+         UnwrapExplicitBrackets: S("[_a] -> _a")
 }
 
 /// Set of unbuilt rules.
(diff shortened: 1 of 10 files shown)


ayazhafiz:
@luke-bhan I'm sorry this is a big PR again 🙁 I thought it would be way simpler, and it was, until I realized I had to do the bootstrapping thing. If you can think of other ways to do this, lmk.

ayazhafiz:
@luke-bhan ptal

libslide/src/evaluator_rules/registry.rs
```diff
@@ -33,8 +33,8 @@ define_rules! {
                AdditiveIdentity: S("_a + 0 -> _a")
                ReorderConstants: S("#a + $b -> $b + #a")
              DistributeNegation: S("-(_a - _b) -> _b - _a")
-                   UnwrapParens: M(&["($a) -> $a", "(#a) -> #a"])
-                   UnwrapBraces: M(&["[$a] -> $a", "[#a] -> #a"])
+                   UnwrapParens: S("(_a) -> _a")
+                   UnwrapBraces: S("[_a] -> _a")
 }
```
luke-bhan, line 38: stop using braces.... they are brackets!

libslide/src/evaluator_rules/registry.rs
```diff
@@ -55,18 +55,31 @@ impl RuleSet {
...
+            //    be represented by a string rule. For example, the rule
+            //
+            //      -(_a - _b) -> _b - _a
+            //
+            //    in its raw form may only be applied to an expression with explicit parentheses.
+            //    By bootstrapping this rule with a rule that removes explicit parentheses, the
+            //    rule can be applied on expressions with implicit parentheses (i.e. of the prefix
+            //    form `(- (- _a _b))`).
+            UnbuiltRule::S(_) => sum + 2,
             _ => sum + 1,
```
luke-bhan, line 63: doesn't this limit our form of rules to just 2 for m set rules. Is there a case where theres more then 2 forms?

libslide/src/grammar.rs
```diff
@@ -133,14 +136,32 @@ impl fmt::Display for Expr {
...
-    Mod,
-    Exp,
+    Plus = 1,
+    Minus = 2,
+    Mult = 10,
+    Div = 11,
+    Mod = 12,
+    Exp = 20,
+}
+
```
luke-bhan, line 148: why did we set defaults?

libslide/src/grammar.rs
```diff
@@ -230,14 +271,28 @@ impl fmt::Display for UnaryOperator {
...
+                let format_arg = |arg: &Rc<$expr>| match arg.as_ref() {
+                    BinaryExpr(l) => format!("({})", l),
+                    expr => expr.to_string(),
+                };
+                result.push_str(&format!("{}{}", self.op, format_arg(&self.rhs)));
+                f.write_str(&result)
+            }
+        }
+    };
 }
```
luke-bhan, line 293: I don;t understand why we are matching with a BinaryExpr in a unary Expr. Can you explain?

luke-bhan:
Can you explain what bootstrapping is? Sorry, I am just not really sure in terms of software.

libslide/src/evaluator_rules/registry.rs
ayazhafiz: I got rid of M rules

libslide/src/grammar.rs
ayazhafiz: So that we can do partial comparison simply. Take a look at the `PartialOrd` impl

libslide/src/grammar.rs
ayazhafiz: A unary expression of the form `(- (+ 1 2))` must be written as `-(1 + 2)`. We match against a bin expr for the above reason. If the argument for the unary expression is a binary expression we have to wrap the binary expression in parentheses, otherwise it's incorrect (would be `-1 + 2`)

libslide/src/grammar.rs
luke-bhan: Ahh i see. I think we should add comments here for that.

libslide/src/grammar.rs
luke-bhan: gotcha

PR #64, opened 2020-05-14
Parse implicit multiplication

ayazhafiz:


libslide/src/parser/test_utils.rs
```diff
@@ -1,22 +1,29 @@
...
-        let tokens = scan($program);
+    ($parser:ident, $inout:expr) => {
+        let inout: Vec<&str> = $inout.split(" => ").collect();
+        let pin = inout[0];
+        let pout = if inout.len() > 1 {
+            inout[1].to_owned()
+        } else {
+            pin.to_owned()
+        };
+        let tokens = scan(pin);
```
luke-bhan, line 11: rewrite if into match for consistency. surprised clippy didn't yell at you for this.

ayazhafiz: Actually, clippy yells if I write it in a match.... wants me to use if/else.

luke-bhan: weird ok.

PR #65, opened 2020-05-14
Apply rules on associative nodes

### LINKED ISSUE #55: Apply rules on associative nodes
Our rules don't match well against associative nodes. For example, `a + 1 + 2` is not reduced any further, though it should become `a + 3`. This is because in the parsed form `((a + 1) + 2)`, of course `a + 1` cannot be reduced, and we can't reduce `1 + 2` since we don't rewrite the expression into `(a + (1 +2))`.

The point is we should rewrite the expression. I think the best way to go about this is to flatten the expression (for example as `+ a 1 2`) and then return a series of unflattened expressions we try to match against (e.g. `+ (a 1) 2`, `+ (a 2) 1`, `+ (1 2) a`)

ayazhafiz:
We do this by flattening binary expressions, rewriting with various
associativities, and running the rewritten expressions on various rules.
If a rewritten expression can be applied by the rule, we use it.
Otherwise we use the rewritten expression whose partial transformation
is the "least complex" (i.e. has the least # of operations).

This still doesn't fix something like `"1 + 2 -b + 3 + a"`. I need to
think more on that. @luke-bhan if you'd like, have a go on it.

Closes #55

### DIFF
--- a/libslide/src/evaluator_rules/registry.rs
+++ b/libslide/src/evaluator_rules/registry.rs
@@ -4,6 +4,7 @@ use super::rule::*;
 use super::unbuilt_rule::UnbuiltRule;
 use fn_rules::*;
 
+use std::cmp::Ordering;
 use std::collections::HashMap;
 
 macro_rules! define_rules {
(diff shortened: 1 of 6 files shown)


libslide/src/evaluator_rules/registry.rs
```diff
@@ -37,6 +38,23 @@ define_rules! {
...
+
+impl Ord for RuleName {
+    fn cmp(&self, other: &Self) -> Ordering {
+        if self == other {
+            return Ordering::Equal;
+        } else if (*self as u8) < (*other as u8) {
+            return Ordering::Less;
+        }
+        Ordering::Greater
+    }
```
luke-bhan, line 55: can't you just do *self as u8.cmp(*other as u8) here?

ayazhafiz: smart...

PR #67, opened 2020-05-15
Validate rules during building

### LINKED ISSUE #66: Validate rules
Validate rule mappings are sane and resolvable before applying them.

ayazhafiz:
Closes #66

### DIFF
--- a/libslide/src/evaluator_rules/registry.rs
+++ b/libslide/src/evaluator_rules/registry.rs
@@ -2,10 +2,13 @@ mod fn_rules;
 
 use super::rule::*;
 use super::unbuilt_rule::UnbuiltRule;
+use crate::utils::indent;
 use fn_rules::*;
 
+use core::fmt;
 use std::cmp::Ordering;
 use std::collections::HashMap;
+use std::error::Error;
 
 macro_rules! define_rules {
     ($($kind:ident: $def:expr)*) => {
@@ -53,6 +56,7 @@ impl Ord for RuleName {
 /// Set of unbuilt rules.
 pub struct RuleSet {
     rules: HashMap<RuleName, UnbuiltRule>,
+    custom_rules: Vec<UnbuiltRule>,
 }
 
 impl Default for RuleSet {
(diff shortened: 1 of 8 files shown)


libslide/src/evaluator_rules/registry.rs
```diff
@@ -64,26 +69,43 @@ impl RuleSet {
...
     pub fn remove(&mut self, rule: RuleName) {
         self.rules.remove(&rule);
     }
 
+    /// Insert a custom unbuilt rule into the rule set.
+    #[allow(unused)] // Used in testing. TODO: enable
+    fn insert_custom<T: Into<UnbuiltRule>>(&mut self, rule: T) {
+        self.custom_rules.push(rule.into());
+    }
+
```
luke-bhan, line 108: is this an error in linter because even functions used in testing do not come up with a warning otherwise like all of ours would.

libslide/src/evaluator_rules/rule.rs
```diff
@@ -1,19 +1,59 @@
...
+            indent("\n", 4),
+            self.map,
+            self.map.from,
+            missing_pats,
+            self.map.to
+        )
+    }
+}
+
+impl Error for UnresolvedMapping {}
```
luke-bhan, line 55: why do we need this?

libslide/src/evaluator_rules/rule.rs
```diff
@@ -57,6 +97,23 @@ impl PatternMap {
...
+        if unresolved_pats.is_empty() {
+            return None;
+        }
+
+        Some(UnresolvedMapping {
+            map: self.clone(),
+            unresolved_pats,
+        })
+    }
 }
```
luke-bhan, line 117: where does this return an error?

ayazhafiz: See the return type and exactly these lines. UnresolvedMap is the error.

libslide/src/evaluator_rules/rule.rs
ayazhafiz: Error is the generic error trait in Rust. This means we can pass it along as an Error in a function that accepts or returns an Error, without having to explicitly know it’s an UnresolvedMapping. Basically the reason you would use a trait (interface).

libslide/src/evaluator_rules/registry.rs
ayazhafiz: Yep it’s a linter (actually rust compiler) warning. Functions only used in testing but available publicly are considered dead code.

PR #68, opened 2020-05-17
Small refactor of evaluation process

ayazhafiz:
I am looking into some way to evaluate expressions like

```
1 + 2 - b + 3 + a
```

To do this, I think the best way forward is to flatten math expressions
with negation rewritten as subtraction. That is, the expression above
should become

```
1 + 2 + -b + 3 + a
```

The code to do this should belong to the `add` rule. As a precursor,
this commit sets up `add` to fold constants in a flattened expression,
and changes symmetric expression pattern matching to only happen in a
string matching context (that is the only place we really care about it,
since a function rule should be able to handle any analysis it needs
itself).

### DIFF
--- a/libslide/src/evaluator_rules/registry/fn_rules.rs
+++ b/libslide/src/evaluator_rules/registry/fn_rules.rs
@@ -1,8 +1,11 @@
 use crate::grammar::*;
+use crate::utils::*;
+
+use std::rc::Rc;
 
 macro_rules! get_binary_args {
     ($expr:expr, $op:pat) => {
-        match $expr {
+        match $expr.as_ref() {
             Expr::BinaryExpr(BinaryExpr { op: $op, lhs, rhs }) => {
                 match (lhs.as_ref(), rhs.as_ref()) {
                     (Expr::Const(l), Expr::Const(r)) => Some((l, r)),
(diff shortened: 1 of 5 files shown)


libslide/src/evaluator_rules/registry/fn_rules.rs
```diff
@@ -14,52 +17,82 @@ macro_rules! get_binary_args {
...
+    let mut i = 0;
+    for _ in 0..args.len() {
+        match args[i].as_ref() {
+            Expr::Const(f) => {
+                konst += f;
+                args.swap_remove(i);
+            }
+            _ => i += 1,
+        }
+    }
```
luke-bhan, line 52: wait so if you match, we don't increment?

libslide/src/utils/grammar.rs
```diff
@@ -48,19 +48,19 @@ fn get_flattened_binary_args(expr: Rc<Expr>, op: BinaryOperator) -> VecDeque<Rc<
             {
                 args.push_front(arg);
             }
-            args.append(&mut get_flattened_binary_args(Rc::clone(&child.rhs), op));
-            args
+            args.extend(&mut get_flattened_binary_args(Rc::clone(&child.rhs), op).drain(..));
+            args.into_iter().collect()
         }
```
luke-bhan, line 53: out of curiosity what does .. op mean? It says "and the rest" on the docs which confused me

ayazhafiz: `drain` takes a range to drain. `..` is a range to the entire container (eq to infinite range)

libslide/src/evaluator_rules/registry/fn_rules.rs
ayazhafiz: Yeah, because if we match then we swap the current index with the last. Then we need to check the current index against to make sure we don't miss anything.

1 x a
^ ---- check 0, add to const

a x
^ ---- we swap removed 0, check a at the same index now

PR #72, opened 2020-05-19
gcd

ayazhafiz:
Part of #71

@luke-bhan I'm sorry for the num-traits dependency... I couldn't figure
out a better way to parametrize the gcd function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/target
+target/
 Cargo.lock
 
 # latex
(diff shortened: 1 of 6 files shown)


luke-bhan:
i wrote an entire comlpex class to avoid the num dependency i hate you <3

ayazhafiz:
> i wrote an entire comlpex class to avoid the num dependency i hate you <3

Well this only adds num-traits, not num ;)

ayazhafiz:
Yep just testing the speeds. I have a feeling for small numbers euclid's will fare far better, but I'll check that after we have a Z polynomial solver

PR #73, opened 2020-05-20
Bignum mul

luke-bhan:
this is broken. I don't know why but a few digits are off. Can you take a look?

### DIFF
--- a/libslide/src/bignum.rs
+++ b/libslide/src/bignum.rs
@@ -2,6 +2,7 @@ use std::fmt;
 mod add;
 mod compare;
 mod complex;
+mod mul;
 mod negate;
 mod sub;
 mod utils;
@@ -57,6 +58,9 @@ impl Bignum {
             } else {
                 return Err(INPUT_ERR_MSG);
             }
+        }
+        if !dec.is_empty() && dec.iter().all(|&i| i == 0) {
+            dec.clear();
         }
         Ok(Bignum {
             is_neg,
--- a/libslide/src/bignum/complex.rs
+++ b/libslide/src/bignum/complex.rs
@@ -1,5 +1,6 @@
 #![allow(clippy::suspicious_arithmetic_impl)]
 #![allow(dead_code)]
+use std::convert::From;
 use std::convert::TryFrom;
 use std::fmt;
 use std::ops;
@@ -7,7 +8,7 @@ use std::ops;
 static INPUT_ERR_MSG: &str = "Input is not valid";
 const TOLERANCE: f64 = 1E-12;
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Default)]
 pub struct Complex {
     real: f64,
     imag: f64,
@@ -30,6 +31,13 @@ impl Complex {
         Complex {
             real: self.real,
             imag: -1.0 * self.imag,
+        }
+    }
+
+    pub fn round(self) -> Complex {
+        Complex {
+            real: self.real.round(),
+            imag: self.imag.round(),
         }
     }
 
@@ -121,6 +129,21 @@ impl TryFrom<String> for Complex {
             Err(_) => return Err(INPUT_ERR_MSG),
         }
         Ok(Complex { real, imag })
+    }
+}
+
+impl From<u8> for Complex {
+    fn from(item: u8) -> Self {
+        Complex {
+            real: item as f64,
+            imag: 0.0,
+        }
+    }
+}
+
+impl From<Complex> for u16 {
+    fn from(item: Complex) -> Self {
+        item.real as u16
     }
 }
 
--- a/libslide/src/bignum/mul.rs
+++ b/libslide/src/bignum/mul.rs
@@ -1,0 +1,106 @@
+#![allow(clippy::suspicious_arithmetic_impl)]
+use crate::bignum::utils::convert_poly;
+use crate::bignum::utils::fft;
+use crate::bignum::utils::ifft;
+use crate::bignum::utils::ispowerof2;
+use crate::bignum::utils::normalize_vecs;
+use crate::bignum::utils::recast_user_vec;
+use crate::bignum::utils::truncate_zeros;
+use crate::bignum::Bignum;
+use std::ops;
+
+impl ops::Mul for Bignum {
+    type Output = Bignum;
+
+    fn mul(self, rhs: Bignum) -> Bignum {
+        let is_neg: bool = self.is_neg ^ rhs.is_neg;
+        // the length of the final size is the sum of all the initial sizes
+        let total_length = self.dec.len() + self.int.len() + rhs.dec.len() + rhs.int.len();
+        // need self and rhs dec lengths for split later
+        // we need to make the vectors the same length at both the decimal and integer level.
+        let (mut lhs_dec, mut rhs_dec) = normalize_vecs(self.dec, rhs.dec);
+        let dec_len = lhs_dec.len() * 2;
+
+        let (mut lhs_int, mut rhs_int) = normalize_vecs(self.int, rhs.int);
+
+        // vector lengths must be a power of 2 to make use of recursive fft
+        while rhs_int.len() + rhs_dec.len() < total_length {
+            rhs_int.push(0);
+            lhs_int.push(0);
+        }
+        while !ispowerof2(rhs_int.len() + rhs_dec.len()) {
+            rhs_int.push(0);
+            lhs_int.push(0);
+        }
+        rhs_dec = rhs_dec.into_iter().rev().collect();
+        lhs_dec = lhs_dec.into_iter().rev().collect();
+        rhs_dec.append(&mut rhs_int);
+        lhs_dec.append(&mut lhs_int);
+        let rhs_len = rhs_dec.len();
+        let rhs_cplx = recast_user_vec(rhs_dec).unwrap();
+        let lhs_cplx = recast_user_vec(lhs_dec).unwrap();
+        let mut rhs_fft = fft(rhs_cplx, rhs_len, false);
+        let lhs_fft = fft(lhs_cplx, rhs_len, false);
+
+        for i in 0..rhs_len {
+            rhs_fft[i] = rhs_fft[i] * lhs_fft[i];
+        }
+
+        let res: Vec<u16> = recast_user_vec(
+            ifft(rhs_fft, rhs_len)
+                .into_iter()
+                .map(|e| e.round())
+                .collect(),
+        )
+        .unwrap();
+        let mut dec_vec: Vec<u8> = convert_poly(res);
+        let mut int_vec: Vec<u8> = dec_vec.split_off(dec_len);
+
+        // remove preceeding zeros
+        dec_vec = dec_vec.into_iter().rev().collect();
+        int_vec = truncate_zeros(int_vec);
+        dec_vec = truncate_zeros(dec_vec);
+
+        Bignum {
+            is_neg,
+            int: int_vec,
+            dec: dec_vec,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    macro_rules! bignum_test_mul {
+        ($($name: ident: $lhs:expr, $rhs:expr, $program:expr)*) => {
+        $(
+            #[test]
+            fn $name() {
+                use crate::bignum::Bignum;
+                let lhs = Bignum::new($lhs.to_string()).unwrap();
+                let rhs = Bignum::new($rhs.to_string()).unwrap();
+                let result = $program.to_string();
+                assert_eq!((lhs*rhs).to_string(), result);
+            }
+        )*
+        }
+    }
+
+    mod mul {
+        bignum_test_mul! {
+            int1: "99", "99", "9801"
+            int2: "5", "5", "25"
+            int3: "10", "10", "100"
+            int4: "999", "999", "998001"
+            int5: "9", "1", "9"
+            int6: "9", "10", "90"
+            int7: "25", "25", "625"
+            int8: "12345678", "12345678", "152415765279684"
+            float1: "0.001", "0.01", "0.00001"
+            float2: "0.003", "0.212", "0.000636"
+            float3: "0.192781230589", "0.12182387511", "0.02348535655882644773979"
+            mixed1: "1.0", "1.0", "1"
+            mixed2: "12912572835.19235098273", "19325812.193812388322389", "249545957551850939.50568665382187490021134197"
+        }
+    }
+}
--- a/libslide/src/bignum/utils.rs
+++ b/libslide/src/bignum/utils.rs
@@ -1,6 +1,11 @@
 #![allow(dead_code)]
 use crate::bignum::complex::Complex;
 use crate::bignum::Bignum;
+use std::cmp::Ordering;
+use std::f64::consts::PI;
+
+const TOLERANCE: f64 = 1E-12;
+static ERR_MSG: &str = "Try From Method failed to cast";
 
 pub fn recast_vec<T, U>(v: Vec<T>) -> Vec<U> {
     let mut v = std::mem::ManuallyDrop::new(v);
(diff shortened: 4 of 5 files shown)


luke-bhan:
It was a fucking casting error between the complex type and the u16 jesus christ 3 hours on that one!

luke-bhan:
let me know. Im sure there's plenty of bad things going on with this one. there are some very weird scenarios I had to handle so if I am missing comments lmk. Also, I am not sure why travis failed? To rebase, I had to fix merge conflicts, but I copied and pasted the entire files from master that had the conflicts so I'm unsure why its failing?

libslide/src/bignum.rs
```diff
@@ -58,6 +70,9 @@ impl Bignum {
                 return Err(INPUT_ERR_MSG);
             }
         }
+        if check_zero(&dec) {
+            dec = vec![];
+        }
```
ayazhafiz, line 75: Actually, you don't even need the check for whether the vector is empty in `is_zero`. Just check if it's all zeros (true if the vector is empty), and then call `clear` on the vector rather than allocating a new one.

libslide/src/bignum/complex.rs
```diff
@@ -33,6 +33,13 @@ impl Complex {
         }
     }
 
+    pub fn round(self) -> Complex {
+        Complex {
+            real: self.real.round(),
+            imag: -1.0 * self.imag,
+        }
+    }
```
ayazhafiz, line 41: What? Shouldn't 2.13 + 4.56i round to 2 + 5i? What's with the imaginary part

libslide/src/bignum/complex.rs
```diff
@@ -124,6 +131,23 @@ impl TryFrom<String> for Complex {
...
+}
+
+impl TryFrom<Complex> for u16 {
+    type Error = &'static str;
+    fn try_from(item: Complex) -> Result<Self, &'static str> {
+        Ok(item.real as u16)
+    }
+}
+
```
ayazhafiz, line 150: Why not just implement `From` since this will always succeed

libslide/src/bignum/utils.rs
```diff
@@ -18,40 +49,42 @@ pub fn abs(rhs: &Bignum) -> Bignum {
     }
 }
 
+pub fn ispowerof2(n: f64) -> bool {
+    if n == 0.0 {
+        return false;
+    }
+    n.log2().ceil() - n.log2().floor() < TOLERANCE
+}
```
ayazhafiz, line 57: Does it have to be an f64? If we can cast to an unsigned int, this is much better:

```
n != 0 && (n & (n-1) == 0)
```

libslide/src/bignum/utils.rs
```diff
@@ -10,6 +13,34 @@ pub fn recast_vec<T, U>(v: Vec<T>) -> Vec<U> {
...
+    Ok(res)
+}
+
+pub fn round_complex_vec(v: Vec<Complex>) -> Vec<Complex> {
+    let mut res = v.clone();
+    for i in 0..v.len() {
+        res[i] = v[i].round()
+    }
+    res
+}
```
ayazhafiz, line 42: You don't have to clone since you took ownership of the vec, just declare v as mut. Anyway, I think it would be better to call into_iter and map wherever this is needed rather than having this util

libslide/src/bignum/utils.rs
```diff
@@ -18,40 +49,42 @@ pub fn abs(rhs: &Bignum) -> Bignum {
...
+        a_even[i] = item[i * 2];
+        a_odd[i] = item[i * 2 + 1];
+    }
+    a_even = fft(a_even, n / 2, ifft);
+    a_odd = fft(a_odd, n / 2, ifft);
+
+    let mut t: Complex;
+    for k in 0..n / 2 {
+        if ifft {
+            t = Complex::new(0.0, 2.0 * std::f64::consts::PI * k as f64 / n as f64).exp() * a_odd[k]
```
ayazhafiz, line 79: Just import PI

ayazhafiz, line 79:
```rust
let coeff = 4 * (ifft as usize) - 2 // 1 => 2, 0 => -2
t = Complex::new(0.0, coeff * ...)
```
this way you don't need to branch

libslide/src/bignum/utils.rs
```diff
@@ -64,6 +97,27 @@ pub fn ifft(item: Vec<Complex>, n: usize) -> Vec<Complex> {
     a
 }
 
+// converts a multi digit polynomial to an single digit vector
```
ayazhafiz, line 100: Can you add an example in the doccomments of this function?

ayazhafiz, line 100: Also
```suggestion
/// converts a multi digit polynomial to an single digit vector
```
this creates a rustdoc comment that will be picked up by our docs creator :)

libslide/src/bignum/utils.rs
```diff
@@ -64,6 +97,27 @@ pub fn ifft(item: Vec<Complex>, n: usize) -> Vec<Complex> {
...
+pub fn convert_poly(v: Vec<u16>) -> Vec<u8> {
+    let mut carry: u16 = 0;
+    let mut a: Vec<u16> = v
+        .into_iter()
+        .enumerate()
+        .map(|(_i, mut e)| {
```
ayazhafiz, line 106: you don't need enumerate...

libslide/src/bignum/utils.rs
```diff
@@ -64,6 +97,27 @@ pub fn ifft(item: Vec<Complex>, n: usize) -> Vec<Complex> {
...
+            e += carry;
+            carry = e / 10;
+            while e / 10 > 0 {
+                e %= 10;
+            }
```
ayazhafiz, line 111: By definition this loop will only run once. So why not just `e %= 10`?

libslide/src/evaluator_rules/registry.rs
```diff
@@ -218,7 +218,6 @@ mod tests {
     (1) Could not resolve pattern map
         "_a -> _b"
     Specifically, source "_a" is missing pattern(s) "_b" present in target "_b"
-
```
ayazhafiz, line 221: this was failing? Thanks :)

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::bignum::Bignum;
+use std::cmp::Ordering;
+use std::ops;
+
+impl ops::Mul for Bignum {
+    type Output = Bignum;
+
+    fn mul(self, rhs: Bignum) -> Bignum {
+        let is_neg: bool = self.is_neg ^ rhs.is_neg;
+        // total length of return vector is p+q+1
```
ayazhafiz, line 17: add comment for what p and q are. Also, 1 is not in `total_length` below.

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,146 @@
...
+        let total_length = self.dec.len() + self.int.len() + rhs.dec.len() + rhs.int.len();
+        // need self and rhs dec lengths for split later
+        let dec_len = self.dec.len() + rhs.dec.len();
+        // we need to make the vectors the same length at both the decimal and integer level.
+        let (mut lhs_dec, mut rhs_dec) = match self.dec.len().cmp(&rhs.dec.len()) {
+            Ordering::Greater => {
+                let mut rhs_dec = vec![0; self.dec.len()];
+                for (i, e) in rhs.dec.iter().enumerate() {
+                    rhs_dec[i] = *e
+                }
```
ayazhafiz, line 27: I think it would be better to reserve `self.dec.len() - rhs.dec.len()` more memory in `rhs.dec` and fill the new memory with 0s.

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,146 @@
...
+                (self.dec, rhs_dec)
+            }
+            Ordering::Less => {
+                let mut lhs_dec = vec![0; rhs.dec.len()];
+                for (i, e) in self.dec.iter().enumerate() {
+                    lhs_dec[i] = *e
+                }
+                (lhs_dec, rhs.dec)
```
ayazhafiz, line 35: ditto

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            Ordering::Equal => (self.dec, rhs.dec),
+        };
+
+        let (mut lhs_int, mut rhs_int) = match self.int.len().cmp(&rhs.int.len()) {
+            Ordering::Greater => {
+                let mut rhs_int = vec![0; self.int.len()];
+                for (i, e) in rhs.int.iter().enumerate() {
+                    rhs_int[i] = *e
+                }
```
ayazhafiz, line 45: ditto.. you get it

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,146 @@
...
+                let mut lhs_int = vec![0; rhs.int.len()];
+                for (i, e) in self.int.iter().enumerate() {
+                    lhs_int[i] = *e
+                }
+                (lhs_int, rhs.int)
+            }
+            Ordering::Equal => (self.int, rhs.int),
+        };
+
+        // need to make vectors mod 2 for fft
```
ayazhafiz, line 58: I think this could be better phrased. Maybe "vector lengths must be a power of 2 for fft"?

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,146 @@
...
+        while rhs_int.len() + rhs_dec.len() < total_length {
+            rhs_int.push(0);
+            lhs_int.push(0);
+        }
+        while !ispowerof2((rhs_int.len() + rhs_dec.len()) as f64) {
+            rhs_int.push(0);
+            lhs_int.push(0);
+        }
+        rhs_dec = rhs_dec.iter().rev().cloned().collect();
+        lhs_dec = lhs_dec.iter().rev().cloned().collect();
```
ayazhafiz, line 68: Use `into_iter`, then there is no need to clone

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,146 @@
...
+        let lhs_fft = fft(lhs_cplx, rhs_len, false);
+
+        for i in 0..rhs_len {
+            rhs_fft[i] = rhs_fft[i] * lhs_fft[i];
+        }
+
+        let res_round = round_complex_vec(ifft(rhs_fft, rhs_len));
+        let res: Vec<u16> = recast_user_vec(res_round).unwrap();
+        let mut dec_vec: Vec<u8> = convert_poly(res);
+        print!("{:?}", dec_vec);
```
ayazhafiz, line 84: remove? or make `debug!` so it's not included in production code

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,146 @@
...
+        if dec_len > 0 {
+            while dec_vec[0] == 0 {
+                dec_vec.remove(0);
+            }
+        }
```
ayazhafiz, line 89: Maybe it's better to use [`skip_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip_while) to avoid shifting down every time

ayazhafiz: On the other hand, do we care if there are trailing zeros? If this is about formatting, we can just add more code in the fmt implementation. Doing this is kind of expensive because we have to shift things around or allocate more, and makes no difference in terms of the math right

libslide/src/bignum/complex.rs
luke-bhan: this is a weird issue I missed! It was written like that but for some reason I guess during merge conflicts it picked up the negate part for imag. Will Fix good eye!

libslide/src/bignum.rs
luke-bhan: smart!

libslide/src/bignum/mul.rs
luke-bhan: I think i just did a neat trick so we can remove them in the end. We still want to remove the zeros because it does slow our processing time down.

luke-bhan:
@ayazhafiz made some changes. Also refactored some code in the sub method since we were reusing. LMK what you think.

libslide/src/bignum/utils.rs
luke-bhan: e can be greater then 100 so it runs more then once

libslide/src/bignum/utils.rs
luke-bhan: so wait  how do I do contribute to our rust docs beyond the triple / or is that the way to do it

libslide/src/bignum/utils.rs
luke-bhan: the log2 function only is available for f64 not usigned

ayazhafiz: Yeah but you can replace the whole function with `n != 0 && (n & (n-1) == 0)`, then you dont need log2

libslide/src/bignum/utils.rs
ayazhafiz: Just triple /. you can also add code blocks in triple backticks (```) and include markdown

libslide/src/bignum/utils.rs
ayazhafiz: Yes, but `e %= 10` will always make `e < 10` right. So then `e/10 > 0` will be false

libslide/src/bignum/mul.rs
ayazhafiz: why does it slow the processing time down?

libslide/src/bignum/utils.rs
```diff
@@ -10,6 +15,56 @@ pub fn recast_vec<T, U>(v: Vec<T>) -> Vec<U> {
...
 
+// note we cannot use the above recast since we have a user defined type with
+// an undefined way of storing its memory
+// prereqs: tryfrom(or from) is implemented for type coercion and default and clone are implemented
+pub fn recast_user_vec<
+    T: std::clone::Clone,
+    U: std::default::Default + std::clone::Clone + std::convert::TryFrom<T>,
+>(
+    v: Vec<T>,
+) -> Result<Vec<U>, &'static str> {
```
ayazhafiz, line 26: nit: more idiomatic to use a where clause
```suggestion
pub fn recast_user_vec<T, U>(v: Vec<T>) -> Result<Vec<U>, &'static str>
where
    T: std::clone::Clone,
    U: std::default::Default + std::clone::Clone + std::convert::TryFrom<T>,
{
```

libslide/src/bignum/utils.rs
```diff
@@ -10,6 +15,56 @@ pub fn recast_vec<T, U>(v: Vec<T>) -> Vec<U> {
...
+            count += 1;
+        } else {
+            break;
+        }
+    }
+    v.truncate(v.len() - count);
+    v
+}
+
+pub fn equalize_vecs(v1: Vec<u8>, v2: Vec<u8>) -> (Vec<u8>, Vec<u8>) {
```
ayazhafiz, line 51:
```suggestion
pub fn normalize_vecs(v1: Vec<u8>, v2: Vec<u8>) -> (Vec<u8>, Vec<u8>) {
```

libslide/src/bignum/mul.rs
luke-bhan: well because if we have extra zeros just sitting in our vector and then we say add it, it still is going to compute through those zeros.

luke-bhan:
Ok, updated the comments. @ayazhafiz

libslide/src/bignum/mul.rs
```diff
@@ -0,0 +1,101 @@
...
+        let lhs_cplx = recast_user_vec(lhs_dec).unwrap();
+        let mut rhs_fft = fft(rhs_cplx, rhs_len, false);
+        let lhs_fft = fft(lhs_cplx, rhs_len, false);
+
+        for i in 0..rhs_len {
+            rhs_fft[i] = rhs_fft[i] * lhs_fft[i];
+        }
+
+        let res: Vec<u16> =
+            recast_user_vec(ifft(rhs_fft, rhs_len).iter().map(|e| e.round()).collect()).unwrap();
```
ayazhafiz, line 50:
```suggestion
            recast_user_vec(ifft(rhs_fft, rhs_len).into_iter().map(|e| e.round()).collect()).unwrap();
```

libslide/src/bignum/utils.rs
ayazhafiz: Please fix this

libslide/src/bignum/utils.rs
ayazhafiz: Also this comment

libslide/src/bignum/utils.rs
```diff
@@ -10,6 +15,55 @@ pub fn recast_vec<T, U>(v: Vec<T>) -> Vec<U> {
...
+{
+    let mut res: Vec<U> = vec![U::default(); v.len()];
+    for i in 0..v.len() {
+        let val = match U::try_from(v[i].clone()) {
+            Ok(val) => val,
+            Err(_err) => return Err(ERR_MSG),
+        };
+        res[i] = val;
+    }
+    Ok(res)
```
ayazhafiz, line 34: Note: this would be more efficient with unsafe code, but we can change that in a later commit.

libslide/src/bignum/utils.rs
```diff
@@ -18,52 +72,64 @@ pub fn abs(rhs: &Bignum) -> Bignum {
...
 
+// converts a multi digit polynomial to an single digit vector
+pub fn convert_poly(v: Vec<u16>) -> Vec<u8> {
+    let mut carry: u16 = 0;
+    let mut a: Vec<u16> = v
+        .into_iter()
+        .map(|mut e| {
+            e += carry;
+            carry = e / 10;
+            e %= 10;
```
ayazhafiz, line 123: Just return e%10

luke-bhan: gotcha

PR #74, opened 2020-05-22
Bignum mod

luke-bhan:
Positive modulo operator for bignum. Rather simple. Built off the mul branch. only file that is affected is modulo.rs

### DIFF
--- a/libslide/src/bignum/modulo.rs
+++ b/libslide/src/bignum/modulo.rs
@@ -1,0 +1,125 @@
+#![allow(clippy::suspicious_arithmetic_impl)]
+use crate::bignum::Bignum;
+use std::ops;
+
+// these functions are used for benchmarks, we currently use binary_skip_mod
+fn binary_skip_mod(lhs: Bignum, rhs: Bignum) -> Bignum {
+    // we only support positive mod for now
+    if !rhs.dec.is_empty() {
+        panic!("The module fuction only supports whole numbers!");
+    }
+    let mut result = lhs;
+    let zero = Bignum::new("0".to_string()).unwrap();
+    let mut coeff = Bignum::new("1".to_string()).unwrap();
+    let multiplier = Bignum::new("2".to_string()).unwrap();
+    while result > zero {
+        result = result - rhs.clone() * coeff.clone();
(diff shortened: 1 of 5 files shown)


libslide/src/bignum/modulo.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn rem(self, rhs: Bignum) -> Bignum {
+        // we only support positive mod for now
+        if !rhs.dec.is_empty() {
+            panic!("The module fuction only supports whole numbers!");
+        }
+        let mut result = self;
+        let zero = Bignum::new("0".to_string()).unwrap();
+        while result >= zero {
+            result = result - rhs.clone();
+        }
```
ayazhafiz, line 17: Can you make the implementation more efficient using binary skipping? First you subtract x, then 2x, then 4x, etc. If you go over you backup and start again from x. Lmk if this makes sense.

Also, only compare with `>`

luke-bhan: yeah I definitely can. never heard of that, but thanks for introducing it to me

luke-bhan: wait do you think it will be faster, because we need to them implement an extra multiply

ayazhafiz: implement both and benchmark :-)

luke-bhan:
how are you organizing benchamrk files?I see you have a separate folder but when I run cargo bench it does not run.

ayazhafiz:
What;s the error?

luke-bhan:
no error, just the benchmark results aren't produced. It only runs all the tests which are ignored.

ayazhafiz:
Make sure you run

cargo bench —features benchmark-internals

Отправлено: Wednesday, May 27, 2020 10:42:53 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Review requested <review_requested@noreply.github.com>
Тема: Re: [ayazhafiz/slide] Bignum mod (#74)


no error, just the benchmark results aren't produced. It only runs all the tests which are ignored.

luke-bhan:
:) thank you

luke-bhan:
binary skipping is faster :)

PR #77, opened 2020-05-28
Eval fn test fail

luke-bhan:
This issue occurs to because the spacing in the test does not consistently translate when you use cargo fmt --all. The solution is just to remove it by formatting the rule errors without a line between them. I think this is fine formatting and saves me the headache of fixing this on every PR.


luke-bhan:
will rebase once bignum mod is merged. Only look at second commit

luke-bhan:
@ayazhafiz this should be rebased!

Issue #78, opened 2020-05-29
Comparison doesn't work when one value is missing a decimal vector


timothycobrien:
For example, `(1.25 > 1)` should return `true`, but it is returning `false` in this case. Similarly, `(1 < 1.25)` should return `true`, but it returns `false`. 

So it seems to occur whenever the floors of the two values are the same, but one of them is an integer (i.e. equal to its own floor). Need to look at `libslide/src/bignum/compare.rs`

timothycobrien:
I'm totally cool with fixing this, was just curious about which solution we might want to go with here? Seeing two main lines of attack.

## Fix 1
BigNums always have at least one digit for both the non-decimal and decimal sides, making things like 2 or .3 impossible as they would be replaced with 2.0 and 0.3. This would mean modifying the BigNum constructor/possible other methods that might omit a side.

## Fix 2
Just fix the comparison so that when a BigNum with both a non-decimal and decimal is compared to a BigNum without either, the comparator knows to treat an empty vector as a zero. Would juts required modifying `compare.rs`

Thoughts @luke-bhan @ayazhafiz ?

luke-bhan:
Go for it!
Interesting, the first fix you have is actually how I had it implemented before the last commit. 
The reason I stopped the ability to hold the zero was that numbers like 2.0 would cause issues with the fft multiplication. I think the second fix is a much easier one and I would probably do it like that since it avoids extra steps in computation long term for both the add and sub method since you don't have to handle the extra position in the decimal vector.

timothycobrien:
I'm wondering, though, if the second fix doesn't impact future code in similar ways. That is to say, all future code would have to account for this same specific case rather than handling all cases.

ayazhafiz:
I think the second fix is better. We shouldn't rely on the bignum to be in any particular state aside from the ordering of its internal containers. If implementations ever change the comparator should still function as long as the underlying data is of the same form.

@timothycobrien do you want me to add you as a collaborator on the project?

timothycobrien:
That'd be great @ayazhafiz

timothycobrien:
I'll fix this now with the modifications to the bignum ctor

PR #79, opened 2020-05-30
fixed compare and added tests

timothycobrien:
Fixing #78 , also modified .gitignore for my IDE

### DIFF
--- a/libslide/src/bignum/compare.rs
+++ b/libslide/src/bignum/compare.rs
@@ -6,6 +6,13 @@ impl Ord for Bignum {
         let greater: Ordering;
         let less: Ordering;
         if self.is_neg ^ other.is_neg {
+            /* need to check for negative 0
+            use the fact that a pure zero will have an empty int and empty dec vector
+            if all the lengths sum to zero, each length must itself be 0
+             */
+            if (self.dec.len() + other.dec.len() + self.int.len() + other.int.len()) == 0 {
+                return Ordering::Equal;
+            }
             return other.is_neg.cmp(&self.is_neg);
         } else if other.is_neg && self.is_neg {
             greater = Ordering::Less;
@@ -41,8 +48,7 @@ impl Ord for Bignum {
                 _ => (),
             }
         }
-
-        match other.dec.len().cmp(&self.dec.len()) {
+        match self.dec.len().cmp(&other.dec.len()) {
             Ordering::Equal => Ordering::Equal,
             Ordering::Less => less,
             Ordering::Greater => greater,
@@ -95,15 +101,23 @@ mod tests {
             equal1: "1 = 1"
             equal2: "1.1 = 1.1"
             equal3: "-25.25 = -25.25"
+            equal4: "1.0 = 1"
+            equal5: "0.0 = 0"
+            equal6: "0 = -0"
+            equal7: "000 = 0"
+            equal8: "000 = 0.0"
+            equal9: "0.000 = 0.0"
             not_equal1: "1 != 2"
             not_equal2: "1.1 != 1.2"
             not_equal3: "2.1 != 1.1"
             not_equal4: "-1.1 != 1.1"
+            not_equal5: "1.1 != 1"
             less_than1: "1 < 2"
             less_than2: "1.1 < 2.1"
             less_than3: "2.1 < 2.2"
             less_than4: "-1 < 1"
             less_than5: "-2 < -1"
+            less_than6: "1 < 1.8"
             less_than_equal1: "1 < 2"
             less_than_equal2: "1.1 <= 2.1"
             less_than_equal3: "2.1 <= 2.2"
@@ -112,11 +126,17 @@ mod tests {
             less_than_equal6: "1 <= 1"
             less_than_equal7: "1.1 <= 1.1"
             less_than_equal8: "-25.25 <= -25.25"
+            less_than_equal9: "1 <= 1.1"
+            less_than_equal10: "1 <= 1.0"
             greater_than1: "2 > 1"
             greater_than2: "2.1 > 1.1"
             greater_than3: "2.2 > 2.1"
             greater_than4: "1 > -1"
             greater_than5: "-1 > -2"
+            greater_than6: "1.2 > 1"
+            greater_than7: "1 > .5"
+            greater_than8: ".5 > .2"
+            greater_than9: ".5 > -1"
             greater_than_equal1: "2 >= 1"
             greater_than_equal2: "2.1 >= 1.1"
             greater_than_equal3: "2.2 >= 2.1"
@@ -125,6 +145,8 @@ mod tests {
             greater_than_equal6: "1 >= 1"
             greater_than_equal7: "1.1 >= 1.1"
             greater_than_equal8: "-25.25 >= -25.25"
+            greater_than_equal9: "1.1 >= 1"
+            greater_than_equal10: "1 >= 1.0"
         }
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ assets/*.log
 assets/*.md5
 assets/*.out
 assets/*.pdf
+
+# Intellij
+.idea*
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,13 @@ jobs:
 
     - name: "Squashed commits"
       if: type = pull_request
-      env: BRANCH="origin/$TRAVIS_PULL_REQUEST_BRANCH"
+      env:
+        - BRANCH="https://github.com/$TRAVIS_REPO_SLUG/$TRAVIS_PULL_REQUEST_BRANCH"
+        - REPO="https://github.com/ayazhafiz/slide"
       script: 
         - git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
         - git fetch origin
-        - scripts/squashed
+        - scripts/squashed REPO BRANCH
 
   allow_failures:
     - rust: nightly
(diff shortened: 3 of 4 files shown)


luke-bhan:
ok, depending on what @ayazhafiz thinks, maybe we should just edit the constructor, I didn't think this would be that kind of problem. I'm sorry. It is just a one liner if we do that so something like what we do with the dec vector:
```
if !int.is_empty() && int.iter().all(|&i| i == 0) { 
    int.clear();
}
```
Then just I would just check compare.rs if all vectors are empty and return equal.
(Also, I think the less than error comes from a switching of orders in the match statement on the last line of compare.rs) 

I think that would shorten this commit to 6 or 7 lines. And I don't think we will run into integration tests since all the operators right now strip the leading zeros which would work for a vector of all zeros. What do you think? @timothycobrien

ayazhafiz:
@luke-bhan do all operations drop trailing zeros on the decimal as well?

Explicitly dropping zeros everywhere is fine, but if we do that let's document that we're doing so on the bignum struct.

libslide/src/bignum/compare.rs
```diff
@@ -3,53 +3,104 @@ use std::cmp::Ordering;
 
 impl Ord for Bignum {
     fn cmp(&self, other: &Bignum) -> Ordering {
+        let (lhs, rhs) = (strip_zeros(self), strip_zeros(other));
+        // lhs and rhs have no leading or trailing (in decimal portion) zeros
```
ayazhafiz, line 7: There's no need to allocate a new bignum; just keep a small int pointing to where leading/trailing zeros (int/decimal) stand.

libslide/src/bignum/compare.rs
```diff
@@ -3,53 +3,104 @@ use std::cmp::Ordering;
...
         let greater: Ordering;
         let less: Ordering;
-        if self.is_neg ^ other.is_neg {
-            return other.is_neg.cmp(&self.is_neg);
-        } else if other.is_neg && self.is_neg {
+        if lhs.is_neg ^ rhs.is_neg { // need to check for negative 0
+            // use the fact that a pure zero will have an empty int and empty dec vector
+            if lhs.dec.is_empty() && rhs.dec.is_empty() && lhs.int.is_empty() && rhs.int.is_empty() {
+                return Ordering::Equal;
+            }
```
ayazhafiz, line 14: then this is equivalent to checking `lhs_dec_term == 0 && rhs_dec_term == lhs_dec_term && ...`

luke-bhan:
theyre supposed to. I’ll add tests in a separate pr

ayazhafiz:
Okay, modifying the ctor sounds good to me

libslide/src/bignum/compare.rs
timothycobrien: Do you mean not modifying either of the Vecs at all and just altering ranges for int-by-int checks?

ayazhafiz:
yes. but luke bhan's idea is probably the way to go

timothycobrien:
> Okay, modifying the ctor sounds good to me

I think it makes sense to @ayazhafiz, unless we're specifically making a distinction between *1* and *1.0*, but I don't think that's the case here? I think it makes sense to just enforce a non-empty int Vec and a non-empty dec Vec, so even *0* would be represented as *0.0*. 

This provides a consistent structure which should make future functions much easier to implement.

ayazhafiz:
I'm not sure I see the benefit of `0.0` vs `.` where the int/dec vectors are empty. Can you elaborate with an example where this makes a difference?

timothycobrien:
I didn't necessarily have a specific issue in mind, but doing something like `0+1` where `0` is actually `.` versus `0.0` *seems* more complicated to me, as that edge case has to be coded in versus being handled like any other case.

timothycobrien:
I just thought of this, but it might make sense to modify all of the comparisons entirely so a comparison of `x > y` amounts to `x - y > 0`,  `x < y` amounts to `y - x > 0`, so we can just test if a number is negative? This is actually where having `0` be represented by empty dec and int vectors might come in handy, but not sure if that alone is worth it imo.

ayazhafiz:
> I just thought of this, but it might make sense to modify all of the comparisons entirely so a comparison of `x > y` amounts to `x - y > 0`, `x < y` amounts to `y - x > 0`, so we can just test if a number is negative?

This would need benchmarking but I my hypothesis is it won't be faster.

timothycobrien:
Having a consistent structure with two non empty vectors seems much better to me for future extensibility than the possibility of one being empty, but I'm having a tough time coming up with a concrete reason why. Thoughts @luke-bhan ?

timothycobrien:
> > I just thought of this, but it might make sense to modify all of the comparisons entirely so a comparison of `x > y` amounts to `x - y > 0`, `x < y` amounts to `y - x > 0`, so we can just test if a number is negative?
> 
> This would need benchmarking but I my hypothesis is it won't be faster.

Was mostly thinking in terms of re-usability here, would also be easier to modify later on (somehow providing a different binary relation than the typical `<, >, =` or modifying their meanings)

luke-bhan:
I see where you're coming from and I don't disagree because it does seem more intuitive. I just don't know how we would handle operations like 0.00 and 0.0000 in the add method. Would the result just be 0.0 so we would have to truncate everything except the last zero? I think its just easier to write a method to trim all vectors (truncate_vectros) and just use that.

ayazhafiz:
I don’t think keeping one zero on both sides is consistent with dropping leading and trailing zeros.

timothycobrien:
Fair point Ayaz, I think it's a matter mostly of convention at this point. Obviously we write `0` on paper, because having `.`  on a paper doesn't make any sense, but no reason why we couldn't just have `.` in the code besides I guess human readability?

ayazhafiz:
yeah, i agree that for the human `0.0` is better. For that the to_string can be modified.

timothycobrien:
This should be good to go now @ayazhafiz, didn't fix the to_string or anything yet, but all the compare tests pass now

libslide/src/bignum/compare.rs
```diff
@@ -6,6 +6,13 @@ impl Ord for Bignum {
         let greater: Ordering;
         let less: Ordering;
         if self.is_neg ^ other.is_neg {
+            /* need to check for negative 0
+            use the fact that a pure zero will have an empty int and empty dec vector
+            if all the lengths sum to zero, each length must itself be 0
```
ayazhafiz, line 11: what if one length is negative ;)

timothycobrien: Troll

timothycobrien:
Any idea on why CI isn't passing @ayazhafiz ?

ayazhafiz:
Shoot, it’s because the squash script expects the branch to be on origin... I’ll fix it after work today, or you can take a go at it

timothycobrien:
> Shoot, it’s because the squash script expects the branch to be on origin... I’ll fix it after work today, or you can take a go at it

I'll take a look

timothycobrien:
Ok nevermind @ayazhafiz, I thought I knew what I was doing vis-a-vis Travis, but I have no idea haha.

Is it not just modifying `.travis.yml` to have this line instead ` env: BRANCH="$TRAVIS_REPO_SLUG/$TRAVIS_PULL_REQUEST_BRANCH"`? Honestly, if you could explain the actual fix to me that would be awesome.

ayazhafiz:
@timothycobrien you're on the right track.

1. You need to get the repo slug and concat it with GitHub url to get a path to a branch. For example
in this case "https://github.com/timothycobrien/slide". Make this a new environment variable or argument for `scripts/squashed`, maybe TARGET_BRANCH_REPO if its an env variable

2. This should be used to add a remote in the same way the `squash_check` remote is set: https://github.com/ayazhafiz/scripts/blob/4fc2ca6217282ef5be9bacb7bcd71e46bfad2781/git/squashed#L29-L30

3. Then, use the branch of the remote in place of the raw BRANCH env variable (https://github.com/ayazhafiz/scripts/blob/4fc2ca6217282ef5be9bacb7bcd71e46bfad2781/git/squashed#L33). Specifically, this should be changed to

```bash
BRANCH="${TARGET_BRANCH_REMOTE}/${BRANCH:-$(git branch --show-current)}"
```

Let me know if this makes sense or if I can help more.

luke-bhan:
@timothycobrien is it ok if I just fix and merge this?

timothycobrien:
I'm looking at it now @luke-bhan, but honestly really new to Travis, so I'll ping you in 30 if I can't make any progress.

timothycobrien:
@ayazhafiz what do the `//` mean on line 33? Also, not quite sure if everything is correct, @luke-bhan please help lol

luke-bhan:
I think you need to change line 37 to
```
BRANCH="${TARGET_BRANCH_REMOTE}/${BRANCH:-$(git branch --show-current)}"
```

timothycobrien:
Still doesn't work @luke-bhan :(

luke-bhan:
oh wait also you need to change the repo env variable to your repo so github/timothycobrien/slide because it is still set to ayazhafiz/slide. There may be other problems but try that.

luke-bhan:
hmm somewhere we have an invalid variable. I would say just echo repo, target_branch_Remote, branch, etc. and it ideally will show where we have the invalid arg

luke-bhan:
instead of dealing with this, I externalized bignum and then updated it. It is now available at https://github.com/luke-bhan/numlib if you are still interested in fixing the fft/implementing new algos. The travis and scripts are not updated yet though. @timothycobrien

timothycobrien:
I'd like to try the other multiplication algo and I have a viable (though hardly complete or optimized) division algo already written but haven't been able to test it without multiplication working correctly.

PR #80, opened 2020-05-30
Bignum test

### LINKED ISSUE #78: Comparison doesn't work when one value is missing a decimal vector

luke-bhan:
This adds tests and removes all leading and trailing zeros in both the int and dec vecs for all operations. It edits the constructor to fix #78, but does not handle the cmp changes required.


luke-bhan:
Jesus idk what I did wrong here, gonna have to fix merge conflicts ignore for now

luke-bhan:
@ayazhafiz should be fixed, lmk.

ayazhafiz:
here is my workflow anytime i want to make a change:

```
git checkout master
git fetch origin # or upstream if you forked into your own repo
git rebase origin/master
git checkout -b my_feature_branch
```

Now your branch is updated with changes from master and there will be no merge conflicts because you have completed a rebase.

luke-bhan:
I think i found my issue. For some reason I had all the logo files still added in my git repo on master. So when I would rebase, they would be added back and obviously they aren't on the remote repo. But, thanks this is helpful.

libslide/src/bignum/modulo.rs
```diff
@@ -36,6 +40,9 @@ fn single_skip_mod(lhs: Bignum, rhs: Bignum) -> Bignum {
     if !rhs.dec.is_empty() {
         panic!("The module fuction only supports whole numbers!");
```
ayazhafiz, line 41:
```suggestion
        panic!("The modulo fuction only supports whole numbers!");
```

can you fix this real quick

libslide/src/bignum/modulo.rs
```diff
@@ -8,6 +8,10 @@ fn binary_skip_mod(lhs: Bignum, rhs: Bignum) -> Bignum {
...
     }
+    print!("{:?}", rhs.int);
```
ayazhafiz, line 11:
```suggestion
```

libslide/src/bignum/add.rs
```diff
@@ -73,6 +74,12 @@ impl ops::Add for Bignum {
         if carry == 1 {
             res_int.push(1);
         }
+        // remove zeros
+        res_int = truncate_zeros(res_int);
+        res_dec = truncate_zeros(res_dec);
+        if res_int.is_empty() && res_dec.is_empty() {
+            is_neg = false;
+        }
```
ayazhafiz, line 82: do we need this?

luke-bhan: it stops the result of -0. We can do a few things here. We can stop it in the to_string method and allow -0 in our code. Or we can remove it in every method.

ayazhafiz: let's just handle it in to string

luke-bhan: ok. will fix.

timothycobrien: -0 is also an issue for comparisons, but we can just handle it in there

libslide/src/bignum.rs
```diff
@@ -65,6 +65,9 @@ impl Bignum {
         if !dec.is_empty() && dec.iter().all(|&i| i == 0) {
             dec.clear();
         }
```
timothycobrien, line 67: Seems like this doesn't remove trailing zeros such as .40000?

libslide/src/bignum.rs
```diff
@@ -65,6 +65,9 @@ impl Bignum {
...
+        if !int.is_empty() && int.iter().all(|&i| i == 0) {
+            int.clear();
+        }
```
timothycobrien, line 70: Won't remove leading zeros like 00004?

luke-bhan: yep good point, will fix;

luke-bhan:
Ok, should be fixed. PTAL @ayazhafiz @timothycobrien

timothycobrien:
Actually one thing I just thought of, do we want to make negative zero a possibility? Or if the int and dec Vecs are both empty, should the is_neg value be set to false as well? Thoughts @ayazhafiz @luke-bhan

No real issue either way? Just need to handle the case in compare `assert(-0==0)`

libslide/src/bignum/modulo.rs
ayazhafiz, line 11: not removed

PR #83, opened 2020-05-31
fixed fft bug

### LINKED ISSUE #82: Multiply doesn't work properly with decimals and integers
Haven't tried all possible test cases, but here is an illustrative one.

`5 * .5` should be equal to `2.5`, but multiply returns the value `5.02`.

@luke-bhan thoughts?

luke-bhan:
Closes #82

### DIFF
--- a/libslide/src/bignum/mul.rs
+++ b/libslide/src/bignum/mul.rs
@@ -28,7 +28,7 @@ impl ops::Mul for Bignum {
             rhs_int.push(0);
             lhs_int.push(0);
         }
-        while !ispowerof2(rhs_int.len() + rhs_dec.len()) {
+        while !ispowerof2(rhs_int.len() + rhs_dec.len()) || rhs_int.len() <= 2 {
             rhs_int.push(0);
             lhs_int.push(0);
         }
@@ -106,6 +106,7 @@ mod tests {
             zero6: "0.0000", "-00000.000", "0"
             zero7: "-0.00", "-0000.00", "0"
             trailing_zero1: "1.00000", "000001.0000", "1"
+            mixed3: "5", ".5", "2.5"
         }
     }
 }


timothycobrien:
These two test cases I added will fail with this code change, but for different reasons.

`zero8: "0.00", "2.23131442121312311", "0"` out of bounds accessing error

`mixed6: "200", ".5", "100"` should obviously be returning 100, but instead returns 0.01? Going to look into this one for now and wait for your thoughts on the "zero8" case @luke-bhan.

timothycobrien:
Looked into things a bit, everything seems fine until line 49 (though I could be wrong), where ifft is called.

At line 49, the vector has contents [Complex { real: 10.0, imag: 0.0 }, Complex { real: -0.0000000000000006123031769111886, imag: 10.0 }, Complex { real: -10.0, imag: 0.0 }, Complex { real: 0.0000000000000006123031769111886, imag: -10.0 }]. Don't know if all of these are correct, but they *seem* reasonable to me. Ifft is too hard to manually debug, so I'm going to wait for your call here @luke-bhan. 

Might also be an issue with the for loop on lines 45-47? And I don't think you want to update dec_len after line 38, but something I noticed and is worth a quick look.

luke-bhan:
Was it this pr that broke those cases I doubt it

timothycobrien:
> Was it this pr that broke those cases I doubt it

Haha nope, can confirm they're broken with/without this PR. But we should probably expand this PR to fix at least the `200*.5 = 0.01` bug and ideally the `"0.00", "2.23131442121312311", "0"` as well, but the latter should be a less common use case.

luke-bhan:
The later case should be an optimization. Same with 1.0, we should just put if statements so it’s faster

libslide/src/bignum/mul.rs
```diff
@@ -28,7 +28,7 @@ impl ops::Mul for Bignum {
             rhs_int.push(0);
             lhs_int.push(0);
         }
-        while !ispowerof2(rhs_int.len() + rhs_dec.len()) {
+        while !ispowerof2(rhs_int.len() + rhs_dec.len()) || rhs_int.len() <= 2 {
```
ayazhafiz, line 31: Why not add the `rhs_dec.len() <= 2` condition as well?

libslide/src/bignum/mul.rs
```diff
@@ -28,7 +28,7 @@ impl ops::Mul for Bignum {
...
             rhs_int.push(0);
             lhs_int.push(0);
```
ayazhafiz, line 33: why do we push to both rhs and lhs

timothycobrien: Preserving even length from what I can see?

timothycobrien:
> The later case should be an optimization. Same with 1.0, we should just put if statements so it’s faster

Definitely for these two cases, curious if the out of bounds error arises for anything else though, gonna investigate now

libslide/src/bignum/mul.rs
luke-bhan: It should the same. This is a mistake.

luke-bhan:
I’ll take a look. The 200*.5 error is just an extension of 5*.5 but I think making the sum will fix this guy

timothycobrien:
> > The later case should be an optimization. Same with 1.0, we should just put if statements so it’s faster
> 
> Definitely for these two cases, curious if the out of bounds error arises for anything else though, gonna investigate now

Well that was quick, `mixed7: "0.01", "4.12564388199914384891937474984", ".0412564388199914384891937474984"` also fails due to out of bounds error. Looking for why that happens now

timothycobrien:
Figured out why it's happening @luke-bhan.

Seems like in line 56, `dec_vec` is capped out at a length of 32, while `dec_len` can easily be greater than that based on line 22 (i.e. in the most recent case `dec_len = 58`)

timothycobrien:
> Figured out why it's happening @luke-bhan.
> 
> Seems like in line 56, `dec_vec` is capped out at a length of 32, while `dec_len` can easily be greater than that based on line 22 (i.e. in the most recent case `dec_len = 58`)

I think 32 might be arbitrary based on the test case, but what I will say is that at some point the 32-length vector full of u16 values needs to be converted to a 61 (?) but at least 58-length vector which is the maximum size of the decimal portion of the test case. That never appears to happen for `recast_user_vec` or `convert_poly`.

luke-bhan:
Ok, that’s prolly not too bad of an issue to fix if you want to rewrite the convert_poly method. I think the superior issue is ordering the .5*200 and cases like that. I’m not sure how to fix that in

timothycobrien:
> Ok, that’s prolly not too bad of an issue to fix if you want to rewrite the convert_poly method. I think the superior issue is ordering the .5*200 and cases like that. I’m not sure how to fix that in

The overall algorithm you're using is Schonhage-Strassen @luke-bhan ? I need a better conceptual understanding of it but once I get that I'll think it over. Might also make sense to switch algorithms if it becomes too difficult to debug?

luke-bhan:
Yeah generally. I mean is just a convolution with b where a and b are the polynomials which Basically means all you do is fft the vector, multiply them in k space, and the ifft back and you’ll have the result in polynomial form

luke-bhan:
Sorry, I am not very good at explaining things. Check out this stack exchange post its part of what I used:
https://math.stackexchange.com/questions/764727/concrete-fft-polynomial-multiplication-example

PR #85, opened 2020-06-02
extern bignum

luke-bhan:
Removes bignum from slide and adds it to https://github.com/luke-bhan/numlib. LMK if I forgot anything. I closed all pull requests and issues with bignum and moved them.


ayazhafiz:
Sgtm. Can you do a follow up PR with two things:

1. Add a bignum dependency in the intermittent so that we can start transitioning to a bignum-like interface in the library. [rug](https://docs.rs/rug/1.9.0/rug/) looks like a good option while we wait on numlib.

2. Add numlib as a dependency in this library (maybe we should do this later though)

Feel free to add me as a reviewer in your numlib PRs, I would be happy to check it out anytime.

luke-bhan:
yes. I wiill make an issue. and you already knew I was definitely making you review those PRs :)

Issue #89, opened 2020-06-04
Rule List


luke-bhan:
This list will serve as a thread for new rules that are to be implemented. This is more of a larger goal todo then a specific issue, but I think organizing our future ideas can't hurt. I will update this list when a new rule idea is posted:
- [ ] handle polynomial factoring x^2 + x / x = x + 1

ayazhafiz:
@luke-bhan there a large number of trivial rules (aka can be represented in the string form “lhs -> rhs”) that we can add. we should figure out some way to aggregate some mathematical identities and add them, or just add these trivial rules as we think of them.

luke-bhan:
I added some to get us started. I think if we were to find and aggregate list that'd be perfect.

ayazhafiz:
https://www.researchgate.net/publication/2450227_A_Complete_Term_Rewriting_System_for_Decimal_Integer_Arithmetic

PR #90, opened 2020-06-04
Add subtractive identity rule

ayazhafiz:
Whether this is equivalent to the additive identity, I'm not sure. Would
love some insight here.

Part of #89


luke-bhan:
hey I’m in bed about to sleep but I’ll give A thorough review tomorrow sorry

ayazhafiz:
you're good, sleep tight!

luke-bhan:
what happens if you try to add the rule a + 1 - 1. Does it simplify to a + 0 or no? Probably not I would guess

ayazhafiz:
yeah it doesnt simplify at all...

PR #93, opened 2020-06-04
Added rug dependency

### LINKED ISSUE #87: Add rug dependency
Lets just use only the rug crate for now. Maybe we will add bignum in the future.

luke-bhan:
closes #87


ayazhafiz:
follow up PR: convert numbers to use this

PR #94, opened 2020-06-04
remove deps from docs

### LINKED ISSUE #88: Emit documentation only for slide and libslide
Currently the docs almost emit documentation for the `num_traits` crate:

https://slide-dev.ayazhafiz.com/num_traits/index.html

We should get rid of this

luke-bhan:
closes #88
Also, we should probably rewrite how we are handling docs, because on travis builds probably isn't a good long term solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
   - scripts/check lintformat
 
 after_success:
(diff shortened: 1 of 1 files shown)


ayazhafiz:
longer-term we can use docs.rs when we publish releases.

Issue #92, opened 2020-06-04
a + 1 - 1 doesn't simplify


ayazhafiz:
Should go to a, goes to a + 1 - 1.

luke-bhan:
I think this could be defined as a rule 
literal - literal  = 0

ayazhafiz:
@luke-bhan I'll assign this to you then

ayazhafiz:
Actually the additive inverse is already a rule so the work that needs to be done here is a little bit different.  I think `a + 1 - 1` is getting parsed as `(- (+ a 1) 1)` and we're currently not smart enough to unroll it as needed. I'm going to readd an S-expression debug form so this is easier to see.

ayazhafiz:
@luke-bhan are you working on this?

luke-bhan:
Not currently. I'm writing the heuristic poly rn so if you wanna do this thats perfect.

ayazhafiz:
sounds good

________________________________
От: Luke Bhan <notifications@github.com>
Отправлено: Saturday, June 6, 2020 3:24:50 PM
Кому: ayazhafiz/slide <slide@noreply.github.com>
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Author <author@noreply.github.com>
Тема: Re: [ayazhafiz/slide] a + 1 - 1 doesn't simplify (#92)


Not currently. I'm writing the heuristic poly rn so if you wanna do this thats perfect.

ayazhafiz:
Leaving this open till I add docs for the commit I made

PR #98, opened 2020-06-06
Support outputting result in s-expression or debug form

### LINKED ISSUE #96: Add S-expression debug formatter for ASTs
We used to have this as the default formatter but got rid of it to make easier pretty-printing. let's add it back.

### LINKED ISSUE #97: Add CLI parser
So that we can pass and handle extra args (for example, stop after parsing or show S-expression form after #96 lands). [clap](https://github.com/clap-rs/clap) is probably the best way to go here.

ayazhafiz:
Also adds an option to dump the AST after parsing.

Closes #96
Closes #97

### DIFF
--- a/slide/src/main.rs
+++ b/slide/src/main.rs
@@ -1,24 +1,83 @@
-use libslide::{evaluate, parse_expression, scan, EvaluatorContext};
+use libslide::{evaluate, parse_expression, scan, EvaluatorContext, Grammar};
 
 use std::env;
 
-fn main() -> Result<(), String> {
-    let program = match env::args().nth(1) {
-        Some(prog) => prog,
-        None => {
-            return Err("Must supply a program.".into());
-        }
-    };
+struct Opts {
+    pub program: String,
+    pub output_form: OutputForm,
+    pub parse_only: bool,
+}
 
-    let tokens = scan(program);
+enum OutputForm {
+    Pretty,
+    SExpression,
+    Debug,
+}
+
+fn get_opts() -> Opts {
+    let matches = clap::App::new(clap::crate_name!())
+        .version(clap::crate_version!())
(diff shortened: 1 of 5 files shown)


ayazhafiz:
Relevant to #92:

```bash
❯ cargo run "a+1-1" -os-expression
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/slide a+1-1 -os-expression`
(- (+ a 1) 1)
```

luke-bhan:
im slightly confused. How do we currently handle this expression in non-debug form?

slide/src/main.rs
```diff
@@ -1,24 +1,83 @@
...
+
+fn print<T>(obj: T, output_form: OutputForm) -> String
+where
+    T: Grammar,
+{
+    match output_form {
+        OutputForm::Pretty => obj.to_string(),
+        OutputForm::SExpression => obj.s_form(),
+        OutputForm::Debug => format!("{:#?}", obj),
+    }
```
luke-bhan, line 82: how is the debug form defined. By obj format implementation which is the same as to_string() right? So why do we have it?

luke-bhan:
other then the debug form, LGTM

ayazhafiz:
Debug form looks like this:

```rust
❯ cargo run "a+1-1" -odebug
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/slide a+1-1 -odebug`
BinaryExpr(
    BinaryExpr {
        op: Minus,
        lhs: BinaryExpr(
            BinaryExpr {
                op: Plus,
                lhs: Var(
                    "a",
                ),
                rhs: Const(
                    1.0,
                ),
            },
        ),
        rhs: Const(
            1.0,
        ),
    },
)
```

PR #99, opened 2020-06-07
Flatten expressions commutatively

### LINKED ISSUE #92: a + 1 - 1 doesn't simplify

ayazhafiz:
This commit adds an option to flatten/fold expressions a little bit smarter
than we currently do by flattening and folding an expression
commutatively.

Not sure if there should be a better name, like `flat_fold` or similar?
Or we can rename the current usages of `flatten` and `unflatten`, which
don't do any expression folding, to `rollup` and `unroll` or similar.
Feedback wanted on this.

Closes #92

### DIFF
--- a/libslide/src/grammar.rs
+++ b/libslide/src/grammar.rs
@@ -294,6 +294,23 @@ pub struct BinaryExpr<E: Expression> {
     pub rhs: Rc<E>,
 }
 
+impl<E> BinaryExpr<E>
+where
+    E: Expression,
+{
+    pub fn mult<T, U>(lhs: T, rhs: U) -> Self
+    where
+        T: Into<Rc<E>>,
+        U: Into<Rc<E>>,
+    {
+        Self {
+            op: BinaryOperator::Mult,
+            lhs: lhs.into(),
+            rhs: rhs.into(),
+        }
+    }
+}
+
 macro_rules! display_binary_expr {
     (<$expr:ident>) => {
         impl fmt::Display for BinaryExpr<$expr> {
@@ -363,6 +380,21 @@ impl fmt::Display for UnaryOperator {
 pub struct UnaryExpr<E: Expression> {
     pub op: UnaryOperator,
     pub rhs: Rc<E>,
+}
+
+impl<E> UnaryExpr<E>
+where
+    E: Expression,
+{
+    pub fn negate<T>(expr: T) -> Self
+    where
+        T: Into<Rc<E>>,
+    {
+        Self {
+            op: UnaryOperator::SignNegative,
+            rhs: expr.into(),
+        }
+    }
 }
 
 macro_rules! display_unary_expr {
--- a/libslide/src/partial_evaluator.rs
+++ b/libslide/src/partial_evaluator.rs
@@ -1,4 +1,6 @@
+mod flatten;
 mod types;
+use flatten::flatten_expr;
 pub use types::*;
 
 use crate::evaluator_rules::RuleSet;
@@ -28,6 +30,9 @@ pub fn evaluate(expr: Stmt, ctxt: EvaluatorContext) -> Result<Expr, Box<dyn Erro
     // meaning we can't simplify any further or are stuck in a cycle.
     let mut expr_hash = hash(&simplified_expr);
     let mut seen: HashSet<u64> = HashSet::new();
+    if ctxt.always_flatten {
+        simplified_expr = flatten_expr(&simplified_expr);
+    }
     while seen.insert(expr_hash) {
         for rule in &built_rules {
             simplified_expr = rule.transform(simplified_expr);
@@ -113,8 +118,7 @@ mod tests {
         subtractive_identity_const:        "1 - 0"       => "1"
         subtractive_identity_any:          "(a * b) - 0" => "a * b"
         subtractive_identity_nested:       "(a + 0) - 0" => "a"
-        // TODO(#89): simplify
-        subtractive_identity_with_reorder: "0 - a - 0"   => "0 - a"
+        subtractive_identity_with_reorder: "0 - a - 0"   => "-a"
 
         reorder_constants:              "1 + a"     => "a + 1"
         reorder_constants_nested:       "1 + a + 2" => "a + 3"
@@ -134,6 +138,8 @@ mod tests {
         unwrap_brackets_nested:           "[a] + [1]" => "a + 1"
 
         flattened_addition:             "1 + 2 - b + 3 - b" => "6 - b - b"
+
+        issue_92: "a + 1 - 1" => "a"
     }
 
     #[test]
--- a/libslide/src/partial_evaluator/flatten.rs
+++ b/libslide/src/partial_evaluator/flatten.rs
@@ -1,0 +1,141 @@
+use crate::grammar::*;
+use crate::utils::{unflatten_binary_expr, UnflattenStrategy};
+
+use std::collections::{HashMap, VecDeque};
+use std::rc::Rc;
+
+pub fn flatten_expr(expr: &Rc<Expr>) -> Rc<Expr> {
+    match expr.as_ref() {
+        Expr::Const(_) | Expr::Var(_) => Rc::clone(expr),
+        Expr::Parend(inner) | Expr::Bracketed(inner) => Rc::clone(inner),
+        Expr::BinaryExpr(BinaryExpr { op, lhs, rhs })
+            if op == &BinaryOperator::Plus || op == &BinaryOperator::Minus =>
+        {
+            flatten_add_or_sub(lhs, rhs, op == &BinaryOperator::Minus)
+        }
+        // TODO: handle everything else
+        _ => Rc::clone(expr),
+    }
+}
+
+pub fn flatten_add_or_sub(o_lhs: &Rc<Expr>, o_rhs: &Rc<Expr>, is_subtract: bool) -> Rc<Expr> {
+    let lhs = flatten_expr(o_lhs);
+    let rhs = flatten_expr(o_rhs);
+
+    let mut coeff = 0.;
+    let mut terms = HashMap::<&Rc<Expr>, f64>::new();
+    let mut args = VecDeque::with_capacity(2);
+    let base_args = [lhs, rhs];
+    args.extend(base_args.iter());
+    let mut args_before_neg = 1;
+    while let Some(arg) = args.pop_front() {
+        let is_neg = is_subtract && args_before_neg == 0;
+        args_before_neg -= 1;
+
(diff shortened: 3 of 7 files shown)


ayazhafiz:
I'll add documentation in a subsequent pass, first want to get feedback if this implementation seems reasonable.

libslide/src/grammar.rs
```diff
@@ -365,6 +382,21 @@ pub struct UnaryExpr<E: Expression> {
...
+where
+    E: Expression,
+{
+    pub fn negate<T>(expr: T) -> Self
+    where
+        T: Into<Rc<E>>,
+    {
+        Self {
+            op: UnaryOperator::SignNegative,
+            rhs: expr.into(),
```
luke-bhan, line 395: shouldn't negate not always be sign negative. if you negating a negative it should flip

libslide/src/partial_evaluator.rs
```diff
@@ -134,6 +138,8 @@ mod tests {
         unwrap_brackets_nested:           "[a] + [1]" => "a + 1"
 
         flattened_addition:             "1 + 2 - b + 3 - b" => "6 - b - b"
+
+        issue_92: "a + 1 - 1" => "a"
     }
```
luke-bhan, line 143: No. We will not be naming these things by issue number- especially rules. Call it literal_cancellation or something like that.

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -0,0 +1,139 @@
+use crate::grammar::*;
+use crate::utils::{unflatten_binary_expr, UnflattenStrategy};
+
+use std::collections::{HashMap, VecDeque};
+use std::rc::Rc;
+
+pub fn flatten_expr(expr: &Rc<Expr>) -> Rc<Expr> {
+    match expr.as_ref() {
```
luke-bhan, line 8: why is expr getting match as_ref. shouldn't it already be ref due to the fact you're borrowing it.

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -0,0 +1,139 @@
...
+        // TODO: handle everything else
+        _ => Rc::clone(expr),
+    }
+}
+
+pub fn flatten_add_or_sub(o_lhs: &Rc<Expr>, o_rhs: &Rc<Expr>, is_subtract: bool) -> Rc<Expr> {
+    let lhs = flatten_expr(o_lhs);
+    let rhs = flatten_expr(o_rhs);
+
+    let mut coeff = 0f64;
```
luke-bhan, line 25: this is a nitpick but we should stay consistnent.. let mut coeff: f64 = 0.0;

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -0,0 +1,139 @@
...
+    let mut terms = HashMap::<&Rc<Expr>, f64>::new();
+    let mut args = VecDeque::with_capacity(2);
+    let base_args = [lhs, rhs];
+    args.extend(base_args.iter());
+    let mut args_before_neg = 1;
```
luke-bhan, line 30: this deque could probably be created in one line;
I think rc should be an empty pointer, but Idk if default is implemented off my head
let mut args: VecDeque = vec![Rc<Expr>::default(), Rc<Expr>::default(), lhs, rhs]

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -0,0 +1,139 @@
...
+            }
+        }
+    }
+
+    let mut new_args: Vec<Rc<Expr>> = Vec::with_capacity(1 + terms.len());
+    if coeff != 0. {
+        new_args.push(Rc::from(Expr::Const(coeff)));
+    }
+    for (term, coeff) in terms.into_iter() {
+        if coeff == 0. {
```
luke-bhan, line 74: you don't need into_iter() here.

libslide/src/partial_evaluator/types.rs
```diff
@@ -2,12 +2,14 @@ use crate::evaluator_rules::RuleName;
...
 }
 
 impl Default for EvaluatorContext {
     fn default() -> Self {
         Self {
             rule_blacklist: vec![],
+            always_flatten: true,
         }
     }
 }
```
luke-bhan, line 15: I think this default constructor should take a parameter to know to flatten rather then set to true but this can be down in another PR.

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -0,0 +1,139 @@
...
+        "1 - 1 -> 0",
+        "1 + 2 - 3 -> 0",
+        "1 - 2 + 3 -> 2",
+        "a - a + 1 -> 1",
+        "a + 1 - 1 -> a",
+    ];
+
+    #[test]
+    fn flatten_cases() {
+        for case in CASES.into_iter() {
```
luke-bhan, line 129: Again, do you need the into_iter() here. I don't think so :)

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -0,0 +1,139 @@
...
+            let mult = BinaryExpr::mult(Expr::Const(coeff), Rc::clone(term));
+            let expr: Expr = mult.into();
+            new_args.push(Rc::from(expr));
+        }
+    }
+
+    match new_args.len() {
+        0 => Rc::from(Expr::Const(0.)),
+        1 => new_args.into_iter().next().unwrap(),
+        _ => unflatten_binary_expr(&new_args, BinaryOperator::Plus, UnflattenStrategy::Left),
```
luke-bhan, line 91: why don't you just use get instead of casting to an iter and unwrapping in new_args

libslide/src/grammar.rs
ayazhafiz: the contructor is just to negate the given expression. I don't think this method should be responsible for unfolding the underlying expression to figure out if it needs to be normalized.

libslide/src/partial_evaluator.rs
ayazhafiz: Why not? These tests are mostly for regressions anyway. If we have a similar issue for something like `a + b + 9 + c -e -9` where 9 - 9 doesn't go to zero, what do we name this? `literal_cancellation_2` is fine but it leaves less context about when and where the bug appeared.

Related to this, I think we should move all tests into some kind of corpus (it can even just be a giant text document), similar to how rust/rustfmt do it. What do you think?

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: This gets a ref to the underlying `Expr`

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: good call, thanks

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: `with_capacity` reserves memory but doesn't populate anything. There is no macro for creating vecdeques in the standard library, but if you want you can roll your own: https://github.com/bluss/maplit/pull/31/files. That can be in a subsequent PR though.

libslide/src/partial_evaluator/types.rs
ayazhafiz: The default ctor takes no parameters because it is the "default". There is a builder method to set this flag in the EvaluatorContext impl.

libslide/src/partial_evaluator/flatten.rs
luke-bhan: No, but you can just make a vector that easily casts into a deque. That’s how they do it on Rust docs

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: `get` returns a reference, but we don't want that, we want to return the value, so it's find to destroy the vec

luke-bhan: Getting a reference and casting it is faster then casting iter element and then getting a next and then unwrapping though

luke-bhan: Getting a reference and casting it is faster then casting iter element and then getting a next and then unwrapping though

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: Can you point me to documentation that does this? I'm not aware of anyway to do this

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: wait what? casting how? We need to go from a reference to a raw value

libslide/src/partial_evaluator/flatten.rs
luke-bhan: https://doc.rust-lang.org/std/collections/struct.VecDeque.html
Reserve function shows them making one from a vec

libslide/src/partial_evaluator/flatten.rs
luke-bhan: I thought you can just dereference

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: hmm..  i personally would rather keep it as is because I don't see a need to create and collect an iterator again. What do you think?

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: No cuz that would move the value out of the vector which is unsound.

libslide/src/partial_evaluator/flatten.rs
luke-bhan: That’s fine

libslide/src/partial_evaluator/flatten.rs
luke-bhan: Why is it unsound if your destroying the deque anyway

ayazhafiz: yeah but the compiler isn't smart enough to know that: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=47cedd7a1acf53f5bff12a0461a41f74

ayazhafiz: There is a better way to do this though, you're right: remove the value from the vector with `swap_remove`

ayazhafiz:
shoot I forgot to add docs!

PR #108, opened 2020-06-11
Add kudos

ayazhafiz:
Thank you @luke-bhan for your work :)


luke-bhan:
Why thank you kind sir

PR #109, opened 2020-06-12
Setup system tests and test harness

### LINKED ISSUE #104: setup a testing framework
I think something similar to rustfmt could be super clean here. For example we would say
"5+5" in the src file and "10" in the target file for an evaluator.

ayazhafiz:
This commit adds a harness for `.slide` system tests, defined under
`slide/src/test`. `.slide` test files have the form

```
!!!args
<CLI args>
!!!args

===in
<program input>
===in

~~~stdout
<standard output>
~~~stdout

~~~stderr
<standard output>
~~~stderr
```

where `!!!args` is an optional clause only needed if the test requires
particular CLI args.

Each test runs through the slide CLI with the specified args and program
input, and the CLI output is checked to match the test-specified stdout
and stderr.

To allow for easy maintainence and updating of tests, tests can be
blessed (have new baselines accepted based off the last test run) using
the new ladder tool:

```
ladder test --sys --bless
```

ladder is a slide development tool that makes it easier to do things
like running system tests. Underneath, the call to run system tests with
a bless is

```
BLESS=1 cargo run -p slide --test slide_tests -- --test_threads=4
```

ladder abstracts this away nicely.

ladder also has a couple other neat features. Running `ladder bench`
will run benchmarks with `--feature=benchmark-internals` passed to cargo
automatically, and `ladder lf` runs `scripts/check lf`.

ladder has some more options; to see them all, try `ladder --help`. I
suggest `alias ladder='./ladder'` so the `./` does not have to be typed.

Closes #104

### DIFF
--- a/ladder
+++ b/ladder
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+
+import argparse
+from argparse import RawTextHelpFormatter
+import os
+import sys
+import subprocess
+
+
+def test(args, runner_args):
+    cargo_args = ["cargo", "test"]
+    env = os.environ.copy()
+    if args.sys:
+        cargo_args.append("--test")
+        cargo_args.append("system_tests")
+    cargo_args.append("--")
+    cargo_args.append("--test-threads")
+    cargo_args.append(str(args.threads))
+    for runner_arg in unknown:
+        cargo_args.append(runner_arg)
+    cargo_args.append("--")
+    if args.filter:
+        cargo_args.append(args.filter)
+
+    if args.bless:
+        env["BLESS"] = "1"
+
+    process = subprocess.run(cargo_args, env=env)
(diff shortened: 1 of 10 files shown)


ladder
```diff
@@ -0,0 +1,132 @@
...
+filter. If no filter is specified, all tests are run.
+
+To run only system tests, pass --sys.
+
+Examples:
+
+> ladder test                  # runs all tests
+> ladder test --sys --green    # runs all system tests and accepts their output
+> ladder test --quiet          # runs `cargo test` in quiet mode
+> ladder test add              # only runs tests with "add" in their name
```
luke-bhan, line 74: I am assuming quiet and green aren't implemented yet or are they a predefined cargo arg

ayazhafiz:
Don't merge yet, I'm going to make this better

Issue #56, opened 2020-05-13
Return parser errors instead of panicking


ayazhafiz:
TSIA

PR #112, opened 2020-06-13
Record token positions during scanning

ayazhafiz:
This will be needed so that we can properly report errors in the lexer
and parser.

Part of #56

### DIFF
--- a/libslide/src/parser.rs
+++ b/libslide/src/parser.rs
@@ -120,7 +120,8 @@ where
         match self.input().peek().map(|t| &t.ty) {
             // <node>(<other>) => <node> * (<other>)
             Some(TokenType::OpenParen) | Some(TokenType::OpenBracket) => {
-                self.input().push_front(Token::new(TokenType::Mult));
+                // TODO: mark this node as synthetic
+                self.input().push_front(Token::new(TokenType::Mult, (0, 0)));
             }
             // <num><var> => <num> * <var>
             Some(TokenType::Variable(_))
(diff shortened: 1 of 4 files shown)


ayazhafiz:
In the parsing phase we lose information about token mapping to original source code without that information being attached to the tokens themselves. So if we want to emit a parse error localized to a token in a source code, we need the source code mapping.

For example:

```
1 + / 2
```

This is clearly unparsable code. We want to error at `/`, but we need to know where `/` is.

timothycobrien:
> In the parsing phase we lose information about token mapping to original source code without that information being attached to the tokens themselves. So if we want to emit a parse error localized to a token in a source code, we need the source code mapping.
> 
> For example:
> 
> ```
> 1 + / 2
> ```
> 
> This is clearly unparsable code. We want to error at `/`, but we need to know where `/` is.

Can't the parser realize that `/` is invalid while parsing and throw the error based on its position? Does it necessarily need to be attached to the token? Only case I can think where it might be is for matching parens.

ayazhafiz:
Yes, but there are two phases of parsing. First we scan the text into a list of tokens, then we parse the tokens. So really, "1 + / 2" is first

```
[1, +, /, 2]
```

and then we try to parse this into an AST (ideally `(+ 1 2)`, but here it fails).

If we did scanning and parsing in one pass we wouldn't need this change, but that's a little bit more complicated than we need right now IMO

ayazhafiz:
Actually I guess scanning/parsing in one go isn't so hard... I've seen TypeScript and Closure do it. But we can save it for later.

timothycobrien:
> Actually I guess scanning/parsing in one go isn't so hard... I've seen TypeScript and Closure do it. But we can save it for later.

Assumed it was one pass, my bad :)

Let's implement it later, simplified version first?

ayazhafiz:
Yep I agree

PR #113, opened 2020-06-13
Add test cases from partial_evaluator

ayazhafiz:
This is a copy of partial_evaluator tests for output and s-expression
cases.


slide/src/test/ui/distribute_negation.slide
```diff
@@ -0,0 +1,15 @@
...
+~~~stdout
+
+~~~stderr
+error: Found argument '-(' which wasn't expected, or isn't valid in this context
+
+USAGE:
+    slide [FLAGS] [OPTIONS] <program>
+
+For more information try --help
+~~~stderr
```
ayazhafiz, line 15: Ahh this is a bug

slide/src/test/ui/negate.slide
```diff
@@ -0,0 +1,15 @@
...
+~~~stdout
+
+~~~stderr
+error: Found argument '-1' which wasn't expected, or isn't valid in this context
+
+USAGE:
+    slide [FLAGS] [OPTIONS] <program>
+
+For more information try --help
+~~~stderr
```
ayazhafiz, line 15: same here

slide/src/test/ui/distribute_negation.slide
ayazhafiz: Going forward, I'm planning to add two more options:
- `~~~exitcode`
- `@TODO: ...` as a directive, with an option to fail on TODO tests

slide/src/test/ui/output_form/s_expression/unwrap_parens_var.slide
```diff
@@ -0,0 +1,19 @@
...
+--parse-only
+-o s-expression
+!!!args
+
+===in
+(a)
+===in
+
+~~~stdout
+(a)
```
timothycobrien, line 11: Shouldn't this just be `a` or am I missing something?

ayazhafiz: We stop at the parsing phase, see `--parse-only` in the args above

PR #116, opened 2020-06-14
Emit nicely-formatted scanning diagnostics

### LINKED ISSUE #102: error reporting
It would be nice (and maybe fun) to have pretty error messages for failures from the slide CLI. For example, for #49 a nice message would be

```
1 - ++++5
    ^^^^^ Hint: this is clearer as "+5"
```

For now I think it's enough to emit such messages at the lexer/parser level<sup>[1](#sup-1)</sup>.

There are a few steps to be done to implement this:

1. Store source position on scanned tokens (easy)
2. Create a module for formatting errors
    - Provide a service for mapping a source byte offset of a token to line/column number (easy-medium)
    - Provide a UI formatter for underlining, wrapping, emitting error messages (medium)

Along with this, a nice test runner with a test UI corpus would likely need to be set up (easy-medium).

---

<sup id="sup-1">1</sup>After parsing we start transforming the source in ways that would be difficult to backpedal and for which errors are not very useful (we can just say the expression is not further reducible). Later on there could be features for which we care about the original source mapping, but IMO that's outside the scope of the project in its current state.

ayazhafiz:
This commit adds a diagnostic module and shows that it behaves well by
emitting some simple scanning errors, like if an invalid token is
encountered.

Closes #102

### DIFF
--- a/libslide/src/diagnostics.rs
+++ b/libslide/src/diagnostics.rs
@@ -1,0 +1,51 @@
+use crate::common::Span;
+
+/// The kind of a slide diagnostic.
+pub enum DiagnosticKind {
+    /// An error diagnostic. Generally, this diagnostic should be emitted for unrecoverable errors.
+    /// In other cases, a warning or a note may be more applicable.
+    Error,
+    /// A note diagnostic is a generic annotation with no specific connotation like `error`. It can
+    /// be particularly useful as an associated diagnostic, for example in expanding on a primary
+    /// error.
+    Note,
+}
+
+/// A diagnostic for slide source code.
+pub struct Diagnostic {
+    pub kind: DiagnosticKind,
+    pub span: Span,
+    pub msg: String,
+    pub associated_diagnostics: Vec<Diagnostic>,
+}
+
+impl Diagnostic {
+    /// Creates an error diagnostic at a span.
+    pub(crate) fn span_err<S, M>(span: S, err: M) -> Diagnostic
+    where
+        S: Into<Span>,
+        M: Into<String>,
+    {
+        Diagnostic {
+            kind: DiagnosticKind::Error,
+            span: span.into(),
+            msg: err.into(),
+            associated_diagnostics: Vec::with_capacity(2),
+        }
+    }
+
+    /// Adds a note to the diagnostic, possibly at a different span.
+    pub(crate) fn with_note<S, M>(mut self, span: S, note: M) -> Diagnostic
+    where
+        S: Into<Span>,
+        M: Into<String>,
+    {
+        self.associated_diagnostics.push(Diagnostic {
+            kind: DiagnosticKind::Note,
+            span: span.into(),
+            msg: note.into(),
+            associated_diagnostics: vec![],
+        });
+        self
+    }
+}
--- a/libslide/src/scanner.rs
+++ b/libslide/src/scanner.rs
@@ -1,18 +1,28 @@
 pub mod types;
 
+use crate::diagnostics::Diagnostic;
 use types::TokenType as TT;
-use types::*;
+pub use types::*;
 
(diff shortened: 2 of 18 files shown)


libslide/src/scanner.rs
```diff
@@ -1,18 +1,28 @@
...
+
+pub fn scan<'a, T: Into<&'a str>>(input: T) -> ScanResult {
+    let mut scanner = Scanner::new(input.into());
     scanner.scan();
-    scanner.output
+    ScanResult {
+        tokens: scanner.output,
+        diagnostics: scanner.diagnostics,
+    }
 }
```
ayazhafiz, line 19: Idk about this interface, it's not exactly the most "rusty" interface but I can't figure out a better way without exposing the Scanner object, which I don't think is either necessary or desirable. LMK

ayazhafiz:
Screen Shot 2020-06-14 at 4 59 09 PM

libslide/src/diagnostics.rs
```diff
@@ -0,0 +1,46 @@
+use crate::common::Span;
+
+/// The kind of a slide diagnostic.
+pub enum DiagnosticKind {
+    Error,
+    Note,
+}
```
luke-bhan, line 7: Whats a note? a warning? I think its worth changing the naming here or some little clarification

libslide/src/scanner.rs
luke-bhan: I think this is a fine way of handling it. Maybe this could be an option wrapped. I'm not sure.

libslide/src/scanner.rs
```diff
@@ -43,6 +54,11 @@ impl Scanner {
         ch
     }
 
+    #[inline]
+    fn push_diag(&mut self, diagnostic: Diagnostic) {
+        self.diagnostics.push(diagnostic);
+    }
+
```
luke-bhan, line 61: Is this function really necessary considering diagnostic is public?

slide/src/diagnostics.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+/// Converts a slide DiagnosticKind to an AnnotationType.
+fn convert_diagnostic_kind(diagnostic_kind: &DiagnosticKind) -> AnnotationType {
+    match diagnostic_kind {
+        DiagnosticKind::Error => AnnotationType::Error,
+        DiagnosticKind::Note => AnnotationType::Note,
+    }
+}
```
luke-bhan, line 67: should these be from functions instead?

ayazhafiz: yeah, but both enums are declared outside of the current crate so rustc won't allow this, sadly ):

libslide/src/diagnostics.rs
ayazhafiz: Good point, I'll add comments

libslide/src/diagnostics.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// In other cases, a warning or a note may be more applicable.
+    Error,
+    /// A note diagnostic is a generic annotation with no specific connotation like `error`. It can
+    /// be particularly useful as an associated diagnostic, for example in expanding on a primary
+    /// error.
+    Note,
+}
+
+/// A diagnostic for slide source code.
+pub struct Diagnostic {
```
timothycobrien, line 15: Should there be multiple kinds of diagnostics that inherit from the same base?

ayazhafiz: There is no inheritance in Rust, only composition. So a diagnostic is one of a set of kinds.

timothycobrien: Oh ok +1 then

PR #120, opened 2020-06-16
Emit parser errors

ayazhafiz:
Part 1 of #56. There are a lot of changes to be made here so I'm going
to split it into several incremental PRs. Sorry for the size of this one
:)

### DIFF
--- a/libslide/src/parser.rs
+++ b/libslide/src/parser.rs
@@ -7,12 +7,13 @@ mod expression_pattern_parser;
 pub use expression_parser::parse as parse_expression;
 pub use expression_pattern_parser::parse as parse_expression_pattern;
 
+use crate::common::Span;
+use crate::diagnostics::Diagnostic;
 use crate::grammar::*;
-use crate::scanner::types::{Token, TokenType};
+use crate::scanner::types::{Token, TokenType as TT};
 use crate::utils::PeekIter;
 
 use core::convert::TryFrom;
-use core::fmt::Display;
 use std::rc::Rc;
 
 macro_rules! binary_expr_parser {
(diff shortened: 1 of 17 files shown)


ayazhafiz:
/slide "1 + "

odds this doesn't work

ayazhafiz:
> slide "1 + " (https://github.com/ayazhafiz/slide/pull/120#issuecomment-644565420)

```
error: Expected an expression, found end of file
 |
1 | 1 + 
 | ^ expected an expression
 |
```

[Command run output][1]

[1]: https://github.com/ayazhafiz/slide-workflows/actions/runs/136862391

libslide/src/parser.rs
```diff
@@ -7,12 +7,13 @@ mod expression_pattern_parser;
 pub use expression_parser::parse as parse_expression;
 pub use expression_pattern_parser::parse as parse_expression_pattern;
 
+use crate::common::Span;
+use crate::diagnostics::Diagnostic;
 use crate::grammar::*;
-use crate::scanner::types::{Token, TokenType};
+use crate::scanner::types::{Token, TokenType as TT};
 use crate::utils::PeekIter;
```
luke-bhan, line 14: can you replace this as TokenType::* and that way you don't need to write tt every time

ayazhafiz: No because the token types overlap with grammar names

luke-bhan: should we fix that or do you think not worth

ayazhafiz: we can fix it but let’s do it in a separate pr

PR #130, opened 2020-06-21
Add design document

ayazhafiz:
@luke-bhan is working on contribution docs, we can link to this design
doc from there. It's also a way for us to keep track of how slide is
structured.


luke-bhan:
I like this. I think we need to explain what an expression pattern is in the parser section. Something like, `an expression pattern represent a rule that the user would like slide to simplify` or something similar

Issue #119, opened 2020-06-15
LGTM gif on LGTM comment


ayazhafiz:
@mahimachander wanted to do this: https://github.com/micnncim/action-lgtm-reaction

ayazhafiz:
@mahimachander will add this

PR #132, opened 2020-06-22
Attach gifs to lgtm comments

mahimachander:
Closes https://github.com/ayazhafiz/slide/issues/119

### DIFF
--- a/.github/workflows/lgtm.yml
+++ b/.github/workflows/lgtm.yml
@@ -1,0 +1,15 @@
+name: Send LGTM reaction
+on:
+  issue_comment:
+    types: [created]
+  pull_request_review:
+    types: [submitted]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


ayazhafiz:
lgtm

.github/workflows/lgtm.yml
```diff
@@ -0,0 +1,15 @@
...
+    types: [submitted]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@1.0.0
+      - uses: micnncim/action-lgtm-reaction@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.SLIDE_WORKFLOWS_TOKEN }}
+          override: true
```
ayazhafiz, line 15: @mahimachander maybe I acted to soon. If you would like, could you fix the override option here? It should not be in the env section, instead it should be in a `with` section, looking something like

```
with:
  overide: true
```

PR #137, opened 2020-06-25
Slide fuzzer

ayazhafiz:
The fuzzer is set to run once a day for 100_000 iterations. If any
failures are found, an issue will be opened by slide-bot. The cron job
will also automatically update fuzz seed files with new slide inputs
present in our integration tests.


ayazhafiz:
Note: by failure, I mean a panic.

luke-bhan:
we're there any significant results from running this

ayazhafiz:
Nope, zero failures after 1M iterations. Which is great

PR #138, opened 2020-06-25
Flatten multiplication and division

### LINKED ISSUE #106: Flatten Multiplication/Division expression
This may not need to be done before 0.1, but I think it would be nice considering adding and subtraction is already implemented and the framework shouldn't take to long.

ayazhafiz:
Also makes some chnages to how we format emitted expressions, since
previously we did it incorrectly. There's a huge amount of documentation
in this PR... I hope that this is a blessing :)

The absolute biggest bet is that this change succeeds in lowering the
expression we present in our readme! So, I think we are ready to cut the
0.1 release after this fix. cc @luke-bhan @timothycobrien

Closes #106


ayazhafiz:
No implementation yet, but I wrote docs first this time so it should way to understand what we're doing (both for reviewers and me). PTAL @luke-bhan @timothycobrien , does this seem correct?

luke-bhan:
LGTM for now. Rerequest review after implementation. Great work explaining this

libslide/src/grammar.rs
```diff
@@ -271,19 +275,27 @@ pub enum BinaryOperator {
...
-        match l.cmp(&r) {
-            Ordering::Less => Some(Ordering::Less),
-            Ordering::Greater => Some(Ordering::Greater),
-            _ => None,
-        }
+impl Ord for BinaryOperator {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let (l, r) = (*self as u8, *other as u8);
+        l.cmp(&r)
     }
```
luke-bhan, line 299: you can't compare directly here?

libslide/src/grammar.rs
```diff
@@ -389,10 +429,47 @@ macro_rules! display_binary_expr {
...
+impl<E> Ord for BinaryExpr<E>
+where
+    E: Expression,
+{
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self.op.cmp(&other.op) {
+            Ordering::Equal => match self.lhs.as_ref().cmp(&other.lhs.as_ref()) {
+                Ordering::Equal => self.rhs.as_ref().cmp(&other.rhs.as_ref()),
+                ord => ord,
+            },
```
luke-bhan, line 450: match within a match here is kind of ugly. can't it just be a single if statement. I know pattern matching is big in rust, but I think doubling up is odd.

libslide/src/grammar.rs
```diff
@@ -389,10 +429,47 @@ macro_rules! display_binary_expr {
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for UnaryOperator {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let (l, r) = (*self as u8, *other as u8);
+        l.cmp(&r)
+    }
```
luke-bhan, line 472: same thing. Can't you compare them straight up.

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -155,6 +172,318 @@ fn flatten_add_or_sub(o_lhs: &Rc<Expr>, o_rhs: &Rc<Expr>, is_subtract: bool) ->
...
+///
+/// And now, all that needs to be done is to construct the flattened expression `2/5 * x^2 / y^-2`.
+fn flatten_mul_or_div(o_lhs: &Rc<Expr>, o_rhs: &Rc<Expr>, is_div: bool) -> Rc<Expr> {
+    let lhs = flatten_expr(o_lhs);
+    let rhs = flatten_expr(o_rhs);
+
+    let mut coeff = 1.;
+    // Term -> # of times it is multiplied. Negative values are equivalent to division.
+    let mut terms = BTreeMap::<&Rc<Expr>, f64>::new();
+
```
luke-bhan, line 383: so why are you using a BTreeMap. I don't see a first or last accesssor here which is my understanding as the main advantage. Wouldn't caching in hashmaps be better

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -155,6 +172,318 @@ fn flatten_add_or_sub(o_lhs: &Rc<Expr>, o_rhs: &Rc<Expr>, is_subtract: bool) ->
...
+                        // 1 * (2 * 3) -> 1 * 2 * 3
+                        args.push_front(lhs);
+                        args.push_front(rhs);
+                        args_before_div += 2;
+                    } else {
+                        // 1 * (2 / 3) -> 1 * 2 / 3; add 2 to the mul side and 3 to the div side.
+                        args.push_front(lhs);
+                        args_before_div += 1;
+                        args.push_back(rhs);
+                    }
```
luke-bhan, line 432: this is clean.

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -155,6 +172,318 @@ fn flatten_add_or_sub(o_lhs: &Rc<Expr>, o_rhs: &Rc<Expr>, is_subtract: bool) ->
...
+                // Otherwise the arg is something we cannot further decompose in this context
+                // (e.g. a variable or an exponentiation), so add it as a term.
+                // TODO: see if we can handle other things more granularly
+                let entry = terms.entry(arg).or_insert(0.);
+                if div_side {
+                    *entry -= 1.;
+                } else {
+                    *entry += 1.;
+                }
+            }
```
luke-bhan, line 445: don't love this part either. I think this todo makes sense

libslide/src/partial_evaluator/flatten.rs
```diff
@@ -155,6 +172,318 @@ fn flatten_add_or_sub(o_lhs: &Rc<Expr>, o_rhs: &Rc<Expr>, is_subtract: bool) ->
...
+    }
+    for (term, coeff) in terms {
+        if coeff == 0. {
+            // The happiest path :)
+            continue;
+        } else if (coeff - 1.).abs() < std::f64::EPSILON {
+            // 1 * x
+            new_args.push(Rc::clone(term));
+        } else if (coeff - -1.).abs() < std::f64::EPSILON {
+            // -1 * x ~ 1/x
```
luke-bhan, line 462: can you explain the double negative here.

luke-bhan:
@ayazhafiz I added comments. I know it's a draft but I saw I was also requested so wasn't sure. Thanks for writing this.

ayazhafiz:
Yeah I should have made it into a full PR. thanks. I’ll make changes soon

Отправлено: Friday, June 26, 2020 8:23:20 PM
Копия: hafiz <ayaz.hafiz.1@gmail.com>; Mention <mention@noreply.github.com>
Тема: Re: [ayazhafiz/slide] Flatten multiplication and division (#138)


@ayazhafiz<https://github.com/ayazhafiz> I added comments. I know it's a draft but I saw I was also requested so wasn't sure. Thanks for writing this.

libslide/src/grammar.rs
ayazhafiz: when he has that genius level IQ 🤤

libslide/src/grammar.rs
ayazhafiz: good point, done

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: HashMap's ordering is non-deterministic, whereas BTreeMap is always sorted. So if we use a HashMap the order of terms in the final expression may not always be the same, which for the system isn't a big a deal but for unit tests matter. LMK if you would prefer I change to a HashMap and call `normalize` during the unit tests.

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: 100% present because of the giant doc block I wrote, it would have been way worse otherwise

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: yeah, the biggest thing is we can better handle multiplications and negations inside of here

libslide/src/partial_evaluator/flatten.rs
luke-bhan: ah I see. I think BTreeMap is fine then bc I'm lazy

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: and powers

libslide/src/partial_evaluator/flatten.rs
ayazhafiz: We are checking if coeff == -1 relative to some epsilon, since we are dealing with machine precision. added a comment

luke-bhan: can't you just write + 1 though

ayazhafiz:
ptal @luke-bhan

libslide/src/partial_evaluator/flatten.rs
luke-bhan: @ayazhafiz

ayazhafiz: yeah but imo it's more clear this way, usually you subtract to compare two items and `-1` is explicit here

ayazhafiz:
ill cut a release tomorrow

PR #155, opened 2020-07-14
Add interner

### LINKED ISSUE #101: interner
Currently we have an implicit interner by common subexpression elimination using `Rc`s (#46). The problem is that an implicit approach makes it very difficult to guarantee when common expressions are shared, and using implementations that depend on these facts (like storing pointer values for expressions) becomes error prone.

I propose we add a dedicated intern module to make it more clear what expressions are interned, and provide an interning guarantee that is abstracted behind the module rather than using `Rc::clone` everywhere. The interner can be a shallow wrapper around `Rc`; in fact, I suggest this is the way to go.

An API I am thinking of would look something like

```rust
pub struct Interner<T> {
    interned: Map<T, Rc<T>>,
}

#[derive(Deref)] // maybe you can't derive this, but it should be implemented
pub struct Interned<T> {
    inner: Rc<T>,
}

impl<T> Interner<T> {
    pub fn intern(&mut self, val: T) -> Interned<T>;
}

impl<T> Interned<T> {
    pub fn copy(&self) -> Self;
}
```

The implementor can flush or change this more to fit the use case in slide, but in general I don't think it should be very complicated.

ayazhafiz:
Closes #101

### DIFF
--- a/libslide/Cargo.toml
+++ b/libslide/Cargo.toml
@@ -26,6 +26,8 @@ required-features = ["benchmark-internals"]
 
 [dependencies]
 strtod = "0.0.1"
+lasso = "0.3.1"
+lazy_static = "1.4.0"
 
 [dependencies.num-traits]
 version = "0.2"
--- a/libslide/src/emit.rs
+++ b/libslide/src/emit.rs
@@ -3,7 +3,6 @@
 use crate::grammar::*;
 
 use core::fmt;
-use std::rc::Rc;
 
 /// The format in which a slide grammar should be emitted.
 #[derive(Copy, Clone)]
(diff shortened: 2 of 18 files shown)


libslide/src/intern.rs
```diff
@@ -0,0 +1,62 @@
...
+    T: Hash + Eq + Clone + Display,
+{
+    pub fn intern(&mut self, val: T) -> Interned<T> {
+        let inner = match self.interned.get(&val) {
+            Some(interned) => interned,
+            None => self.interned.entry(val.clone()).or_insert(val),
+        };
+        Interned { inner }
+    }
+}
```
lukebhan, line 62: Maybe, I dont understand interners, but do we need both an interned, and intern, and an interner classes here? Seems like an over complication

ayazhafiz:
@lukebhan I hate this, I think I am doing too much work for what might be 0 payout. I will hold off on this until our bug bash/refactor.

PR #156, opened 2020-07-14
Feature/ui

lukebhan:
WIP: Right now just have both the app interface in go and the slide backend communicating. @ayazhafiz Can we possibly discuss what we want to do for arguments right now. I would like it to just be like Wolfram and we apply a bunch of predetermined rules at first. Thoughts?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,4 +3,5 @@
 members = [
     "libslide",
     "slide",
+	"app",
 ]
--- a/webapp/index.html
+++ b/webapp/index.html
@@ -1,0 +1,155 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta name="viewport" content="width=device-width" />
+    <meta charset="UTF-8" />
+    <link
+      rel="stylesheet"
+      href="https://unpkg.com/@primer/css/dist/primer.css"
+    />
+    <link
+      rel="stylesheet"
+      href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.0.0/styles/github-gist.min.css"
+    />
+
+    <style>
+      .markdown-body {
+        margin: 0 auto;
+      }
+
+      .options {
+        display: flex;
+        flex-wrap: wrap;
+      }
+
+      .options > div {
+        margin-right: 30px;
+      }
+
+      input:disabled + label {
+        opacity: 0.5;
+      }
+    </style>
+
+    <script src="https://cdn.jsdelivr.net/npm/vue@2.6.10/dist/vue.js"></script>
+    <script
+      src="https://cdnjs.cloudflare.com/ajax/libs/lz-string/1.4.4/lz-string.min.js"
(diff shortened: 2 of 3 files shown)


ayazhafiz:
> I would like it to just be like Wolfram and we apply a bunch of predetermined rules at first.

I think this is a good idea. We can make it more complicated later on; right now it can just run `./slide <program>`

ayazhafiz:
BTW we should think about versioning too. How will people run different versions of slide/what versions do we want to allow? I am thinking all releases (currently just 0.0.1) and the master branch.

lukebhan:
Eventually, I think our UI will be very thorough. Like possible have a set of default rules you can check and uncheck as well as writing your custom rules. However, I think the average person probably doesn't care what version of slide they are using in the UI, thats much more of a developer problem which I think is why we have the slide-bot/tests and can add it there.

lukebhan:
Also, we probably should just switch from golang to like a react/angular frontend for scalability. Thoughts?

ayazhafiz:
> should just switch from golang to like a react/angular frontend

Go is that backend; that can stay. On the frontend that sounds good, but maybe we should use something like Vue so we can have just a single HTML page.

ayazhafiz:
I see, so currently the view is generated by the server. You can also make a static webpage that will communicate to the server without having to refresh. Lmk if you want to video chat about how to do this, maybe for a live stream on your YT ;)

lukebhan:
Yeah, thats exactly what I mean. I am just unsure how do I run the slide application from a go server? maybe we can chat about this bc I understand setting up the api.

app/src/api/main.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+async fn index(info: web::Json<Info>) -> Result<String> {
+    slide(info.input)
+}
+
+#[actix_rt::main]
+async fn main() -> std::io::Result<()> {
+    HttpServer::new(|| App::new().route("/", web::post().to(index)))
+        .bind("127.0.0.1:8088")?.run().await
```
timothycobrien, line 20: This should probably be an environment variable

lukebhan:
I think I will just use commands following something like this tutorial now that I explore more.Thanks. 
https://blog.kowalczyk.info/article/wOYk/advanced-command-execution-in-go-with-osexec.html

ayazhafiz:
I recommend keeping the go server, and getting rid of the rust one.

lukebhan:
Frontend pref @ayazhafiz @timothycobrien react or angular are my choices. This probably will be something we all work on so lmk if you have a pref.

ayazhafiz:
I prefer Vue > React > Angular for this project because I would rather keep everything in a single HTML file. However I am not opposed to a larger frontend app with many sources, it is just more maintenance.

lukebhan:
I haven't used vue before, but that is fine. The intro docs look readable so I'm gonna use that. Hoping to have a small little setup that we can expand on.

ayazhafiz:
dope, im very excited for this feature. thanks for doing this

ayazhafiz:
@lukebhan any updates on this?

PR #157, opened 2020-07-22
Add algorithm for converting decimal to fraction

ayazhafiz:
This commit adds an algorithm for converting a decimal number to its
fraction representation using a convergent algorithm that walks the
[Stern–Brocot tree](https://en.wikipedia.org/wiki/Stern%E2%80%93Brocot_tree).

If the algorithm fails to find an exact fractional representation after
a specified number of iterations, it bails out. This is to allow the
user to decide at which point they want to stop searching for a
fraction.

### DIFF
--- a/libslide/src/math/fraction.rs
+++ b/libslide/src/math/fraction.rs
@@ -1,0 +1,157 @@
+#![allow(unused)]
+
+/// Represents a fraction, consisting of a numerator and denominator.
+type Fraction = (/* num */ i64, /* den */ u64);
+
+#[derive(PartialEq, Debug)]
+pub struct Dec2FracError {
+    num_iter: u64,
+    decimal_error: f64,
+}
+
+impl std::fmt::Display for Dec2FracError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Failed to find an exact fraction representation after {} iterations.\
+            The error at this precision is {}.",
+            self.num_iter, self.decimal_error,
+        )
+    }
+}
+
+/// Converts a decimal number to its irreducible fractional representation by walking the
+/// [Stern-Brocot tree](https://en.wikipedia.org/wiki/Stern%E2%80%93Brocot_tree). This is
+/// equivalent to a binary search of the [Farey sequence](https://en.wikipedia.org/wiki/Farey_sequence).
+///
+/// ## Algorithm
+///
+/// Say we are looking for a number A, where A = Whole + Decimal, Whole ∈ W, and Decimal ∈ [0, 1).
+/// Then if we can find T and B such that Decimal = T / B,
+///      A = Whole + Decimal = (Whole * B) / B + (T / B)
+///        = (Whole * B + T) / B
+///
+/// To find T and B we perform a binary search of irreducible fractions starting with the range [0, 1].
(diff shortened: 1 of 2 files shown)


libslide/src/math/fraction.rs
```diff
@@ -0,0 +1,133 @@
...
+/// ```
+///
+/// We stop when the mediant is equivalent to the decimal searched for.
+///
+/// ## Failure
+///
+/// This algorithm is iterative and convergent, but terminates after `max_iter` iterations. If a
+/// fractional representation equivalent to the decimal at floating point precision cannot be found
+/// after `max_iter`, nothing is returned.
+///
```
lukebhan, line 48: very nice commenting

libslide/src/math/fraction.rs
```diff
@@ -0,0 +1,133 @@
...
+/// ## Examples
+///
+/// ```ignore
+/// assert_eq!(dec2frac(0.5, 10), Some((1, 2)))
+/// assert_eq!(dec2frac(3.14159265358979323, 10), None)
+/// assert_eq!(dec2frac(-3.142857142857142857, 100_000), Some((-22, 7)))
+/// ```
+pub fn dec2frac(mut num: f64, max_iter: u64) -> Option<Fraction> {
+    let coeff = if num < 0. { -1 } else { 1 };
```
lukebhan, line 57: use result instead of option.

ayazhafiz: what would be the error type of the result?

lukebhan: it should explain to the user why the program failed. Returning none is confusing and I don't know where my expression failed

lukebhan: Err("Value {x} could not be converted to decmial")

ayazhafiz: Does this look good?

libslide/src/math/fraction.rs
```diff
@@ -0,0 +1,157 @@
...
+///
+/// ## Algorithm
+///
+/// Say we are looking for a number A, where A = Whole + Decimal, Whole ∈ W, and Decimal ∈ [0, 1).
+/// Then if we can find T and B such that Decimal = T / B,
+///      A = Whole + Decimal = (Whole * B) / B + (T / B)
+///        = (Whole * B + T) / B
+///
+/// To find T and B we perform a binary search of irreducible fractions starting with the range [0, 1].
+/// On each iteration of the search, we compute the mediant of the range.
```
timothycobrien, line 35: I thought this was a fake word

libslide/src/math/fraction.rs
```diff
@@ -0,0 +1,157 @@
...
+    let mut lo = (0, 1);
+    let mut hi = (1, 1);
+
+    fn mediant(lo: (u64, u64), hi: (u64, u64)) -> (u64, u64) {
+        (lo.0 + hi.0, lo.1 + hi.1)
+    }
+
+    let frac = {
+        || {
+            if decimal == 0. {
```
timothycobrien, line 89: What is this dot

libslide/src/math/fraction.rs
ayazhafiz: it kinda is

libslide/src/math/fraction.rs
ayazhafiz: to force floating point :~

Issue #162, opened 2020-07-28
Latex Emit


ayazhafiz:
We should add `latex` emit mode. This emit mode would generate LaTeX code from the slide IR; for
example,

```slide
(x + 1) / (x + 2)
```

would become

```latex
\frac{\left\(x + 1\right\)}{\left\(x + 2\right\)}
```

PR #163, opened 2020-08-05
Add format emit strategies to libslide

ayazhafiz:
Currently, we determine on how a slide invocation should be emitted on
the application side, routing to `fmt::Display`, `fmt::Debug`, or
`Grammar#s_expr` accordingly. This won't scale very well, and in truth
our emission strategies should be the basis for other core rust traits
(like the `fmt` ones) implemented for the libslide IR.

This commit adds an `emit` module that provides an `EmitFormat` enum
replacing the existing `OutputForm` enum on the application side, and
exposes an `Emit` trait used to describe how a type can be emitted.
Eventually all types implementing `Emit` should implement all of its
member functions, but to ease migration, by default only `emit_pretty`
must be implemented.

The nice thing about this structure is that `fmt::Display` comes for
free if your type implements `emit_pretty`, and `emit_debug` comes for
free if your type implements `fmt::Debug`. Eventually, I would like to
reconcile the coupling between the `fmt` macros and `Emit` entirely,
maybe using a proc macro, but that is outside the scope of this change
and requires some forethought.

This work prepares for a LaTeX emit strategy (also low priority, but if
a new contributor would like to add it, this commit lays the groundwork
to make it easier).

Part of #162

### DIFF
--- a/libslide/src/emit.rs
+++ b/libslide/src/emit.rs
@@ -1,0 +1,287 @@
+//! Emit strategies for the libslide grammar IR.
+
+use crate::grammar::*;
+
+use core::fmt;
+use std::rc::Rc;
+
+/// The format in which a slide grammar should be emitted.
+#[derive(Copy, Clone)]
+pub enum EmitFormat {
+    /// Canonical, human-readable form.
+    /// For example, `1+1` is output as `1 + 1`.
+    Pretty,
+    /// S-expression form.
+    /// For example, `1+1` is output as `(+ 1 1)`.
+    SExpression,
+    /// LaTeX output form.
+    /// For example, `(1 + 1)` is output as `\left\(1 + 1\right\)`.
+    /// NB: this is not yet implemented.
+    Latex,
+    /// Slide internal debug form.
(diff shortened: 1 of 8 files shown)


ayazhafiz:
@lukebhan ptal

ayazhafiz:
or @timothycobrien

libslide/src/emit.rs
```diff
@@ -0,0 +1,287 @@
...
+            Self::Parend(inner) => format!("({})", inner.emit_s_expression()),
+            Self::Bracketed(inner) => format!("[{}]", inner.emit_s_expression()),
+        }
+    }
+}
+
+impl Emit for Rc<ExprPat> {
+    fn emit_pretty(&self) -> String {
+        self.as_ref().emit_pretty()
+    }
```
lukebhan, line 282: is this a double refernce? because your pulling &self. I'm a little confused as why you need as_ref() again

ayazhafiz: not great API, `as_ref` is to take the reference of the item held within the `Rc`

PR #171, opened 2020-08-21
Slide webapp UI

### LINKED ISSUE #115: Build a simple UI for handling rules and expressions
I think we should start thinking about an interface for our platform. Some small website where you can add rules (Set a default set of rules) with just checking boxes and then type your expression similar to wolfram and we simplify. This certainly can be on the backlog, but thoughts @ayazhafiz 
@timothycobrien

ayazhafiz:
This continues @lukebhan's work from #156. Closes #115.

### DIFF
--- a/webapp/main.go
+++ b/webapp/main.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"html/template"
+	"net/http"
+)
+
+func buildSlide() {
+	// Clone, build slide.
+	// This should run in a green thread every hour or so?
+	// Builds will be incremental after every pull. Built artifacts should be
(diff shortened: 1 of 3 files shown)


webapp/main.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+
+		tmpl.Execute(w, struct {
+			Success  bool
+			Response string
+		}{true, "dummy"})
+
+	})
+	http.ListenAndServe(":8080", nil)
+}
```
ayazhafiz, line 35: This is still WIP, the server and UI don't communicate yet.

PR #175, opened 2020-08-24
Use \mod for mod emit in latex and wrap latex emit as inline math

ayazhafiz:
cc @timothycobrien

### DIFF
--- a/libslide/src/emit.rs
+++ b/libslide/src/emit.rs
@@ -43,11 +43,15 @@ where
     Self: fmt::Display + fmt::Debug,
 {
     /// Emit `self` with the given [EmitFormat][EmitFormat].
+    ///
+    /// NB: This is a multiplexer of the corresponding `emit_` methods present on [Emit][Emit],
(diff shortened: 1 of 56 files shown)


ayazhafiz:
done

timothycobrien:
@ayazhafiz I believe there shouldn't be a `\` before characters such as `(` or `)` (see file [here](https://github.com/yslide/slide/blob/842a531a81beac8bf5ae5808060bbbe687737cbd/slide/src/test/ui/emit/latex/additive_identity_any.slide))

ayazhafiz:
you're right. want to send a patch?

timothycobrien:
Will do

timothycobrien:
Associated exponents are not correct, how should we go about modifying them? @ayazhafiz 

`$2 ^ 3 ^ 2$` errors, for example

ayazhafiz:
Yes, I noticed that too. They will need to be wrapped in braces according to the association. So that example should be `$2 ^ {3 ^ 2}$`.

PR #178, opened 2020-08-25
Add interner

### LINKED ISSUE #101: interner

ayazhafiz:
This commit adds an interner for slide Grammars. See
libslide/src/grammar/intern.rs for more details. The purpose of this is
to reduce duplication of grammar instances that already exist by
aggregating all instances in a common pool, and returning an opaque
reference to the instance rather than the instance directly. Operations
on the instance can still be done through the intern reference via
dereferences or methods on traits common to Grammars.

The intern strategy is

```
Grammar -> String Hash -> 32-bit identifier
^^^^^^^                   ^^^^^^^^^^^^^^^^^
instance                 interned reference
```

The 32-bit identifier is then mapped back to a Grammar when we need to
dereference it.

Closes #101

### DIFF
--- a/libslide/src/grammar/intern.rs
+++ b/libslide/src/grammar/intern.rs
@@ -1,0 +1,212 @@
+//! Provides interners for slide [Grammar]s.
+//! An interned slide [Grammar] is itself a [Grammar] for convenience of use.
+//!
+//! [Grammar]: super::Grammar
+
+use crate::emit::Emit;
+use crate::grammar::{Expr, ExprPat, Grammar};
+
+use core::cmp::Ordering;
+use lasso::{Rodeo, Spur};
+use lazy_static::lazy_static;
+use std::collections::HashMap;
+use std::ops::Deref;
+use std::sync::RwLock;
+
+/// Describes an interned slide expression.
+pub trait InternedExpression
+where
+    Self: Copy + Deref + Ord + From<super::BinaryExpr<Self>> + From<super::UnaryExpr<Self>>,
+{
+    type Inner;
+
+    /// Returns whether the expression is a statically-evaluatable constant.
+    fn is_const(&self) -> bool;
+
+    /// Paranthesizes `inner`.
+    fn paren(inner: Self) -> Self;
+
+    /// Brackets `inner`.
+    fn bracket(inner: Self) -> Self;
+
+    /// Returns an empty expression.
+    fn empty() -> Self;
+}
+
+macro_rules! make_interner {
+    ($($intern_macro:ident, $ty:ty, $interned_struct:ident, $rodeo:ident, $spur2expr:ident)*) => {$(
+        lazy_static! {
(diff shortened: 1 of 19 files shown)


lukebhan:
Looks great, couldn't find any errors.

libslide/src/grammar/intern.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        /// Interns an expression.
+        #[doc(hidden)]
+        #[macro_export]
+        macro_rules! $intern_macro {
+            ($expr: expr) => {
+                $interned_struct::intern($expr)
+            }
+        }
+
```
lukebhan, line 94: I don't think this is neccessary.

libslide/src/grammar/intern.rs
```diff
@@ -0,0 +1,212 @@
...
+        impl core::fmt::Display for $interned_struct {
+            fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                write!(f, "{}", self.emit_pretty(),)
+            }
+        }
+
+        impl AsRef<$ty> for $interned_struct {
+            fn as_ref(&self) -> &$ty {
+                self.deref()
+            }
```
lukebhan, line 120: why is it deref? is self. here a double pointer?

libslide/src/grammar/intern.rs
```diff
@@ -0,0 +1,212 @@
...
+                    None => {
+                        let spur = $rodeo
+                            .write()
+                            .expect("Failed to write to intern rodeo; likely poisoned.")
+                            .get_or_intern(expr.emit_s_expression());
+                        $spur2expr
+                            .write()
+                            .expect("Failed to write to intern reverse map; likely poisoned.")
+                            .insert(spur, expr);
+                        Self(spur)
```
lukebhan, line 80: i dont understand this.

PR #190, opened 2020-08-30
Support evaluating assignment statements

### LINKED ISSUE #186: Fuzzing failure
Daily slide fuzzing failed. Last 30 lines of the output log are as follows:

```
Reproduce with:

	cargo fuzz run slide fuzz/artifacts/slide/slow-unit-3c88a97ff1d6557c235c51d7b4160c65253365cd

Minimize test case with:

	cargo fuzz tmin slide fuzz/artifacts/slide/slow-unit-3c88a97ff1d6557c235c51d7b4160c65253365cd


────────────────────────────────────────────────────────────────────────────────

Failing input:

	fuzz/artifacts/slide/crash-e7240724ac567b0b5a9b96eaff1dd411e9647f0e

Output of `std::fmt::Debug`:

	"xar = (1 + 10 / [2])"

Reproduce with:

	cargo fuzz run slide fuzz/artifacts/slide/crash-e7240724ac567b0b5a9b96eaff1dd411e9647f0e

Minimize test case with:

	cargo fuzz tmin slide fuzz/artifacts/slide/crash-e7240724ac567b0b5a9b96eaff1dd411e9647f0e

────────────────────────────────────────────────────────────────────────────────

Error: Fuzz target exited with exit code: 77
```

Download this test case at https://transfer.sh/3FZYc/hello.txt

[Command run output][1]

[1]: https://github.com/yslide/slide-workflows/actions/runs/229769563

ayazhafiz:
The idea is we move to evaluating statements in one pass, not just
expressions.

Closes #186

### DIFF
--- a/libslide/src/partial_evaluator.rs
+++ b/libslide/src/partial_evaluator.rs
@@ -5,28 +5,36 @@ mod types;
 use flatten::flatten_expr;
 pub use types::*;
 
-use crate::evaluator_rules::RuleSet;
+use crate::evaluator_rules::{Rule, RuleSet};
 use crate::grammar::*;
 use crate::utils::{hash, normalize};
 
 use std::collections::HashSet;
 use std::error::Error;
 
-/// Evaluates an expression to as simplified a form as possible.
+/// Evaluates a statement to as simplified a form as possible.
 /// The evaluation may be partial, as some values (like variables) may be unknown.
-pub fn evaluate(expr: Stmt, ctxt: EvaluatorContext) -> Result<InternedExpr, Box<dyn Error>> {
+pub fn evaluate(expr: Stmt, ctxt: &EvaluatorContext) -> Result<Stmt, Box<dyn Error>> {
     let mut rule_set = RuleSet::default();
-    for rule in ctxt.rule_denylist {
+    for rule in &ctxt.rule_denylist {
         rule_set.remove(rule)
     }
     let built_rules = rule_set.build()?;
 
-    let mut simplified_expr: InternedExpr = match expr {
-        Stmt::Expr(expr) => expr,
-        // TODO: see below
-        _ => todo!("Evaluation currently only handles expressions"),
-    };
+    Ok(match expr {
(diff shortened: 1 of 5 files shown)


libslide/src/partial_evaluator.rs
```diff
@@ -5,28 +5,36 @@ mod types;
...
+            var,
+            rhs: evaluate_expr(expr, &built_rules, &ctxt),
+        }),
+    })
+}
 
+/// Evaluates an expression to as simplified a form as possible.
+/// The evaluation may be partial, as some values (like variables) may be unknown.
+/// The returned expression is [normalized](crate::utils::grammar::normalize).
+fn evaluate_expr(expr: InternedExpr, rules: &Vec<Rule>, ctxt: &EvaluatorContext) -> InternedExpr {
```
lukebhan, line 36: We should probably pass a slice instead of a &Vec. https://www.reddit.com/r/rust/comments/8kujd1/newbie_correct_way_to_pass_vector_of_references/

libslide/src/partial_evaluator.rs
```diff
@@ -35,13 +43,13 @@ pub fn evaluate(expr: Stmt, ctxt: EvaluatorContext) -> Result<InternedExpr, Box<
...
     while seen.insert(expr_hash) {
-        for rule in &built_rules {
+        for rule in rules {
             simplified_expr = rule.transform(simplified_expr);
         }
         expr_hash = hash(&simplified_expr);
     }
 
-    Ok(normalize(simplified_expr))
+    normalize(simplified_expr)
```
lukebhan, line 52: why are we not using results for the evaluating expressions, but we use a result for evaluate. Seems inconsistent

libslide/src/partial_evaluator.rs
ayazhafiz: good point, done

libslide/src/partial_evaluator.rs
ayazhafiz: because evaluate can fail (see rule_set.build()?) but evaluate_expr will always succeed. Why wrap it in a result if we know it always works?

PR #195, opened 2020-09-01
Add fuzz output to emitted issue and fix workflow url

ayazhafiz:


ayazhafiz:
It uploads a file and return a link to it.

Issue #135, opened 2020-06-23
Linting and multi-phase parsing


ayazhafiz:
@luke-bhan has a good suggestion to add a lint warning that an expression like `++++5` is trivially reducible to `+5`. This lint warning has to be generated after the program is already parsed; today this is not possible because when the program AST is generated, token spans are lost and duplicate AST nodes are eliminated. The latter point means we cannot even keep a registry of AST nodes to their spans.

Going forward I suggest we split the slide IR into at least two phases:

1. Generation of a rich AST, with spans and other niceties, that is suitable for linting and other static evaluation where the original source code is required.

2. When the expression is ready to be run through the partial evaluator, it is lowered to an IR that is equivalent to what we have today and has the same elimination (interning) properties.

Before we do this a formal interner should be built (#101).

<small>This mistake is so common in compiler design, I really hoped it wouldn't matter for this project but I was wrong.</small>

ayazhafiz:
Feel free to assign urself

PR #204, opened 2020-09-08
Add support for spans on expressions

ayazhafiz:
This prepares for a "linter" pass that will walk a parse tree and detect
"bad" patterns, like `++++5` which is trivially reducible to `5`.

Future work may also use these spans to provide intellegent suggestions
on reducibility, for example noting that `3 + 4` is the span that was
converted to `7` in an expression like `2 * 3 / (3 + 4)`.

Also, this change has nearly convinced me that having the interner in a
state as we do now is a useless and bad idea, since it's (1) already
pretty complicated and (2) making expressions unique by their span
defeats the usefulness of caching structurally identical expressions.

Part of #135

### DIFF
--- a/libslide/src/common.rs
+++ b/libslide/src/common.rs
@@ -3,12 +3,30 @@
 /// Describes the character span of a substring in a text.
 ///
 /// For example, in "abcdef", "bcd" has the span (1, 4).
-#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+#[derive(PartialEq, Eq, Copy, Clone, Debug, Hash)]
 pub struct Span {
     /// Inclusive lower bound index of the span
     pub lo: usize,
     /// Exclusive upper bound index of the span
     pub hi: usize,
+}
+
+/// A dummy span for use in places where a span is not (yet) known.
+///
+/// Clearly this span is incorrect since lo = 10001 > 1 = hi, but a well-formed span must observe
+/// the invariant lo <= hi.
+///
+/// NB: This is only to be used during migration and refactoring. Do *not* use this for new
+/// interned expressions.
+pub(crate) static DUMMY_SP: Span = Span { lo: 10001, hi: 1 };
+
+impl Span {
+    pub(crate) fn to(&self, other: Span) -> Span {
+        Self {
+            lo: self.lo,
+            hi: other.hi,
+        }
+    }
 }
 
 impl From<(usize, usize)> for Span {
--- a/libslide/src/grammar/intern.rs
+++ b/libslide/src/grammar/intern.rs
@@ -5,6 +5,7 @@
 
 use crate::emit::Emit;
 use crate::grammar::{Expr, ExprPat, Grammar};
+use crate::Span;
 
 use core::cmp::Ordering;
 use lasso::{Rodeo, Spur};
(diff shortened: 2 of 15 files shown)


lukebhan:
@ayazhafiz is this a wip

ayazhafiz:
No it is ready

ayazhafiz:
The interner finally works... omg

ayazhafiz:
By the way, this is my third attempt at trying to implement spans.

The first attempt was to keep an expression -> span mapping. This didn't get far because I didn't like how brittle it was.

The second attempt was to introduce a `SpannedExpr` that has an `InternedExpr` and a `Span`. The idea is if I want to make a binary expression, and I have an LHS and RHS `SpannedExpr`, I just use their spans to create a new BinaryExpression with a span covering the LHS and RHS, and use that to wrap the BinaryExpression in another SpannedExpr. Then I realized that there is no way to get from the BinaryExpression to the span of the LHS, since the BinaryExpr will only contain a reference to the LHS expression, not its wrapped `SpannedExpr`!

This one works though, and I think it is the best way forward. But probably the interner should be removed ):

libslide/src/common.rs
```diff
@@ -3,14 +3,28 @@
...
     pub lo: usize,
     /// Exclusive upper bound index of the span
     pub hi: usize,
 }
 
+/// A dummy span for use in places where a span is not (yet) known.
+/// Clearly this span is incorrect since lo = 10001 > 1 = hi, but a well-formed span must observe
+/// the invariant lo <= hi.
+pub(crate) static DUMMY_SP: Span = Span { lo: 10001, hi: 1 };
+
```
lukebhan, line 18: this seems bad.

lukebhan: I don't understand this I think, can you explain. I see where you use it for the tests, but don't you think writing the macro to take an option here could be a better route and then if none is passed, you can just use it. Basically were like emulating optional params in rust.

libslide/src/grammar/intern.rs
```diff
@@ -131,15 +204,36 @@ macro_rules! make_interner {
...
+                //   |   |   |   | k |   |     (before reallocation)
+                //   |   |   |   | v |   |
+                //                 \
+                //          /-------\-> expr -- return a reference to this
+                //         /
+                //   |   | v |   |   |   |
+                //   |   | k |   |   |   |     (after reallocation)
+                //
+                //   so even if the map is reallocated, our reference is valid.
                 unsafe {
```
lukebhan, line 235: can you elaborate just for me. I don't understand what you mean by we are returning a ref to an expression. Does that just mean dereferencing? If it is, then i overthought.

libslide/src/common.rs
ayazhafiz: I think it is better to be explicit, otherwise it will look like we are passing "no" span to an expression, even though we always have to have one. Also this way it is easier to find all missing spans since we can grep for DUMMY_SP. I will add a comment that this should not be used in general outside of refactoring. Lmk what you think.

libslide/src/grammar/intern.rs
ayazhafiz: Updated the comment, lmk what you think.

libslide/src/common.rs
lukebhan: that sounds good

PR #208, opened 2020-09-09
Emit inf as latex \infty

ayazhafiz:


libslide/src/emit.rs
```diff
@@ -187,7 +187,11 @@ impl Emit for Expr {
 
     fn emit_latex(&self) -> String {
         match self {
-            Self::Const(num) => num.to_string(),
+            Self::Const(num) => match num.to_string().as_ref() {
+                "inf" => "\\infty",
+                other => other,
```
lukebhan, line 192: why double \\

lukebhan (approved):

libslide/src/emit.rs
ayazhafiz: `\` is an escape character, so we have to use `\\` to escape it (it becomes `\`)

PR #211, opened 2020-09-10
Add help menu test

ayazhafiz:


slide/src/main.rs
```diff
@@ -15,7 +15,10 @@ fn get_opts(color: bool) -> Opts {
         )
         .arg(
             clap::Arg::with_name("output-form")
-                .short("o")
+                .short("-o")
+                .long("--output-form")
+                .next_line_help(true)
+                .help("Slide emit format.")
```
lukebhan, line 21:
```suggestion
                .help("Slide emit format. Acceptable formats are: pretty, s-expression, latex, and debug")
```

lukebhan (approved):

slide/src/main.rs
ayazhafiz: Updated, lmk what you think

lukebhan: very nice.

PR #233, opened 2020-09-16
Support := as an assignment operator

ayazhafiz:
Often := is used as a definition operator and = used as an equality
operator in mathematical expressions. This commit adds support for := in
assignment contexts.

We will also emit a nice diagnostic for cases like `a : = 5`, where a
user probably meant `a := 5`.

### DIFF
--- a/libslide/src/emit.rs
+++ b/libslide/src/emit.rs
@@ -41,6 +41,9 @@ bitflags::bitflags! {
         /// Emit divisions as fractions.
         /// Applies to LaTeX emit.
         const FRAC = 1;
+        /// Emit assignment operators as ":=".
+        /// Applies to pretty emit.
+        const DEFINE_ASSIGN = 2;
     }
 }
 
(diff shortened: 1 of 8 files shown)


ayazhafiz:
@lukebhan I know you are busy can you PTAL

libslide/src/emit.rs
```diff
@@ -41,6 +41,9 @@ bitflags::bitflags! {
         /// Emit divisions as fractions.
         /// Applies to LaTeX emit.
         const FRAC = 1;
+        /// Emit assignment operators as ":=".
+        /// Applies to pretty emit.
+        const DEFINE_ASSIGN = 2;
     }
```
lukebhan, line 47: NIT: change FRAC to define_frac for consistency with these vars

ayazhafiz: well := is the "definition assignment" operator

lukebhan: whats the diff between assign and definition assignment. I don't think is a universal notation for this op. Like we see it in so many ways.

lukebhan (approved):

lukebhan:
this is fine. Not too much code here. The let assign = if... is not necessary but I think it improves readability. the space linter is cool, but itd be cooler if you had :k= or a single char between as a possible look as well.

ayazhafiz:
I agree that would be very nice. But it is very difficult to do and probably impractical.

PR #235, opened 2020-09-18
Support multiple statements across new lines

ayazhafiz:
This commit adds support for parsing and evaluating multiple statements
across lines. This means

```
1 + 2
a = 3
```

will be parsed as two statements correctly. However,

```
1 + 2 a = 3
```

will not, and neither will expression patterns.

### DIFF
--- a/libslide/src/emit.rs
+++ b/libslide/src/emit.rs
@@ -148,6 +148,29 @@ macro_rules! latex_wrap {
     };
 }
 
+#[inline]
+fn join_emits<'a, E: 'a + Emit>(
+    list: impl Iterator<Item = &'a E>,
+    emit: impl FnMut(&E) -> String,
+) -> String {
+    list.map(emit).collect::<Vec<_>>().join("\n")
+}
+
+fmt_emit_impl!(StmtList);
+impl Emit for StmtList {
+    fn emit_pretty(&self, config: EmitConfig) -> String {
+        join_emits(self.iter(), |s| s.emit_pretty(config))
+    }
+
+    fn emit_s_expression(&self, config: EmitConfig) -> String {
+        join_emits(self.iter(), |s| s.emit_s_expression(config))
+    }
+
+    fn emit_latex(&self, config: EmitConfig) -> String {
+        join_emits(self.iter(), |s| s.emit_latex(config))
+    }
+}
+
 fmt_emit_impl!(Stmt);
 impl Emit for Stmt {
     fn emit_pretty(&self, config: EmitConfig) -> String {
--- a/libslide/src/grammar/statement.rs
+++ b/libslide/src/grammar/statement.rs
@@ -1,5 +1,47 @@
 use super::*;
 
+/// A list of statements in a slide program.
+#[derive(Clone, Debug)]
+pub struct StmtList {
+    /// The list of statements.
+    list: Vec<Stmt>,
+}
+
+impl Grammar for StmtList {}
+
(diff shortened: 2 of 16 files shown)


lukebhan:
looking through now

libslide/src/emit.rs
```diff
@@ -148,6 +148,25 @@ macro_rules! latex_wrap {
...
 
+fn join_emits<E: Emit>(list: &[E], emit: impl FnMut(&E) -> String) -> String {
+    list.iter().map(emit).collect::<Vec<_>>().join("\n")
+}
+
+fmt_emit_impl!(StmtList);
+impl Emit for StmtList {
+    fn emit_pretty(&self, config: EmitConfig) -> String {
+        join_emits(&self.list, |s| s.emit_pretty(config))
+    }
```
lukebhan, line 159: can you explain the closure here. and why your passing it to a map? I just understand it

libslide/src/grammar/statement.rs
```diff
@@ -1,5 +1,20 @@
...
+}
+
+impl Grammar for StmtList {}
+
+impl StmtList {
+    pub(crate) fn new(list: Vec<Stmt>) -> Self {
+        Self { list }
+    }
+}
+
```
lukebhan, line 17: youre lazy :)

libslide/src/parser/expression_pattern_parser.rs
```diff
@@ -63,6 +65,10 @@ impl Parser<InternedExprPat> for ExpressionPatternParser {
     fn parse_any_pattern(&mut self, name: String, span: Span) -> Self::Expr {
         intern_expr_pat!(ExprPat::AnyPat(name), span)
     }
+
+    fn has_stmt_break(&mut self) -> bool {
+        false
+    }
 }
```
lukebhan, line 72: why do you need a function here if it always ret false

slide/src/test/ui/extra_items.slide
```diff
@@ -12,6 +12,7 @@ error[P0001]: Unexpected extra tokens
   |
 1 |   1 + 0 -1 2 3
   |  __________^
+  |            - help: if you meant to specify another statement, add a newline before this token
 2 | |   4 5 6 / 7 ^ 8
```
lukebhan, line 16: i don't think these tests are consitent for example. line 14 has a bunch of ___ before the ^ , but the test in tokens_past_expr has 3 ^ values and no _

lukebhan (commented):
@ayazhafiz looked through. ASked some questions and I think the tests are inconsistent.

libslide/src/emit.rs
ayazhafiz: The StmtList is a list of statements. We are passing a closure to describe how each statement should be emitted, then mapping each statement with that emit fn and joining into a string..

libslide/src/grammar/statement.rs
ayazhafiz: I'm smart

libslide/src/parser/expression_pattern_parser.rs
ayazhafiz: Because declared has_stmt_break on the parser trait

slide/src/test/ui/extra_items.slide
ayazhafiz: because the diagnostic in this test spans to line 3. Does that make sense?

PR #244, opened 2020-09-22
Expand variables in expressions with variable definitions

### LINKED ISSUE #243: Expand variables
To support simplifying a series of statements, we need to be able to expand variables in an expression with a variable definition we get elsewhere.

ayazhafiz:
This prepares for a component in the partial evalautor that will
substitute terminated (or nearly-terminated) variable definitions in
expressions using those variables to simplify the expression further.
For example, if we have

```
b = 1 + a + 2 + a
a = 3
```

we currently reduce this to

```
b = 3 + 2a
a = 3
```

but of course we might want to do

```
b = 9
a = 3
```

Closes #243

### DIFF
--- a/libslide/src/grammar/transformer.rs
+++ b/libslide/src/grammar/transformer.rs
@@ -1,7 +1,67 @@
-use super::Grammar;
+use super::*;
+use crate::Span;
 
 /// A trait for transforming one grammar into another.
 /// This transformer takes ownership of the grammar it transforms.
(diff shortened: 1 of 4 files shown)


ayazhafiz:
@lukebhan ptal

lukebhan (approved):
why do we need both lazy and eager?

ayazhafiz:
For now I dont think it hurts to have both

PR #246, opened 2020-09-22
Record span of assignment operator

ayazhafiz:
Needed for a linter I will add to emit warnings when "=" and ":=" are
mixed.

ptal @luke-bhan

### DIFF
--- a/libslide/src/emit.rs
+++ b/libslide/src/emit.rs
@@ -210,9 +210,11 @@ impl Emit for Stmt {
     fn emit_s_expression(&self, config: EmitConfig) -> String {
         match self {
             Self::Expr(expr) => expr.emit_s_expression(config),
-            Self::Assignment(Assignment { var, rhs }) => {
-                format!("(= {} {})", var, rhs.emit_s_expression(config))
-            }
+            Self::Assignment(Assignment {
+                var,
+                rhs,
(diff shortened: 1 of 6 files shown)


lukebhan:
So, then we are going back from what you had wrote with using the = and := as switches of each other depending on the user param and just defining this right?

lukebhan (approved):

ayazhafiz:
What do you think? Right now they will still be emitted as only = or only :=, but we can remove that option.

lukebhan:
I think we should be explicit on those operators. So only one can be used in a program and leave it up to the users. (aka i like the way we have it, I was just making sure that was the intention)

ayazhafiz:
Hm I guess I'm confused. Right now if you do

```
a = 1
b := 1
```

it will be accepted but emit as

```
a=1
b=1
```

or the same with all :=. Is this okay? I plan to add a lint after this commit that will warn when = and := are mixed.

lukebhan:
Oh ok, I thought the linter was already there. I don't think thats ok without saying something to the user, but if you write the linter then its fine, because otherwise maybe the user has defined one as const def and the other as def and thus that could confused them without a warning.

ayazhafiz:
sounds good

PR #252, opened 2020-09-23
Add linter for mixed use of assignment operators

ayazhafiz:


ayazhafiz:
@lukebhan you will like this :)

libslide/src/linter/stmt/homogenous_assignment.rs
```diff
@@ -0,0 +1,76 @@
...
+    ///a = 1
+    ///b := 1
+    ///```
+    ///
+    ///This can be misleading or confusing, as these two operators are syntactically different (and
+    ///semantically different in canonical mathematics notation), but are treated the same in slide.
+    ///
+    ///For this reason, it is suggested that exclusively `=` or `:=` are used for assignments in
+    ///slide programs.
+    L0004: HomogenousAssignmentLinter
```
lukebhan, line 17: what is this: L0004?

lukebhan (approved):
this is so clean. I love how easy it is to write diagonistic with your span.

libslide/src/linter/stmt/homogenous_assignment.rs
ayazhafiz: it is the diagnostic code

PR #255, opened 2020-09-23
Closes #254

### LINKED ISSUE #254: Fuzzing failure
Daily slide fuzzing failed. Last 30 lines of the output log are as follows:

```
Reproduce with:

	cargo fuzz run slide fuzz/artifacts/slide/slow-unit-2095f5ad77126bb98aa179eea098430eba108616

Minimize test case with:

	cargo fuzz tmin slide fuzz/artifacts/slide/slow-unit-2095f5ad77126bb98aa179eea098430eba108616


────────────────────────────────────────────────────────────────────────────────

Failing input:

	fuzz/artifacts/slide/crash-78d3124a761cc25997a65a9d3060e985a128d26a

Output of `std::fmt::Debug`:

	"1 +  *ц"

Reproduce with:

	cargo fuzz run slide fuzz/artifacts/slide/crash-78d3124a761cc25997a65a9d3060e985a128d26a

Minimize test case with:

	cargo fuzz tmin slide fuzz/artifacts/slide/crash-78d3124a761cc25997a65a9d3060e985a128d26a

────────────────────────────────────────────────────────────────────────────────

Error: Fuzz target exited with exit code: 77
```

Download this test case at https://transfer.sh/13YG8B/hello.txt

[Command run output][1]

[Fuzz output file][2]

[1]: https://github.com/yslide/slide/actions/runs/269246306
[2]: https://transfer.sh/FPltL/fuzz-output.txt

ayazhafiz:


slide/src/test/common.rs
```diff
@@ -43,7 +43,7 @@ macro_rules! printer {
 macro_rules! fail {
     ($report:expr) => {
         libtest_mimic::Outcome::Failed {
-            msg: Some(atomic_lock($report)),
+            msg: Some(std::sync::Arc::new(std::sync::Mutex::new($report))),
```
lukebhan, line 46: so youre making an arc to disallow mutating and the using a mutex to mutate? I'm confused by this

slide/src/test/exec.rs
```diff
@@ -27,12 +27,13 @@ pub fn run_slide(args: &str, input: &str) -> Result<SlideOutput, Outcome> {
...
-
-    Ok((stdout, stderr, code.to_string()))
+    match std::panic::catch_unwind(|| slide::run_slide(opts)) {
+        Ok(slide::SlideResult {
+            stdout,
+            stderr,
+            code,
+            ..
+        }) => Ok((stdout, stderr, code.to_string())),
+        Err(_) => Err(print_fail! { Failure: "Test panicked!"; }),
```
lukebhan, line 37: shouldn't we show the error.

lukebhan (approved):
couple comments.

slide/src/test/common.rs
ayazhafiz: No, this is an atomic mutex. So we can pass the mutex between threads.

slide/src/test/exec.rs
ayazhafiz: Ideally yes, but actually there is no way to do so because the err here does not implement display..

PR #256, opened 2020-09-24
Remove expression interner and add string interner

ayazhafiz:
ptal @lukebhan

### DIFF
--- a/libslide/src/utils/grammar.rs
+++ b/libslide/src/utils/grammar.rs
@@ -2,20 +2,20 @@ use crate::grammar::*;
 
 use std::collections::{HashSet, VecDeque};
 
-pub fn get_symmetric_expressions(expr: InternedExpr) -> Vec<InternedExpr> {
+pub fn get_symmetric_expressions(expr: RcExpr) -> Vec<RcExpr> {
     match expr.as_ref() {
         Expr::BinaryExpr(BinaryExpr { op, .. }) => match op {
             BinaryOperator::Plus | BinaryOperator::Mult => {
-                let args = get_flattened_binary_args(expr, *op);
+                let args = get_flattened_binary_args(expr.clone(), *op);
                 let mut sym_args = Vec::with_capacity(args.len());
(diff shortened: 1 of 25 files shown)


libslide/src/utils/grammar.rs
```diff
@@ -224,33 +230,35 @@ pub fn unique_pats<'a>(expr: &'a InternedExprPat) -> HashSet<&'a InternedExprPat
...
-                rhs: normalize(*rhs),
+                rhs: normalize(rhs.clone()),
             }),
             expr.span
         ),
         Expr::Parend(expr) => {
-            let inner = normalize(*expr);
-            intern_expr!(Expr::Parend(inner), inner.span)
+            let inner = normalize(expr.clone());
+            let span = inner.span;
```
lukebhan, line 255: why make an extra var here.

libslide/src/utils/grammar.rs
```diff
@@ -224,33 +230,35 @@ pub fn unique_pats<'a>(expr: &'a InternedExprPat) -> HashSet<&'a InternedExprPat
...
+            rc_expr!(Expr::Parend(inner), span)
         }
         Expr::Bracketed(expr) => {
-            let inner = normalize(*expr);
-            intern_expr!(Expr::Bracketed(inner), inner.span)
+            let inner = normalize(expr.clone());
+            let span = inner.span;
+            rc_expr!(Expr::Bracketed(inner), span)
```
lukebhan, line 261: same as above

lukebhan (approved):

libslide/src/utils/grammar.rs
ayazhafiz: because otherwise we have to clone inner since its used on the next line

libslide/src/utils/grammar.rs
ayazhafiz: because otherwise we have to clone inner since its used on the next line

PR #262, opened 2020-09-27
Add support for arbitrary-precision numbers

ayazhafiz:
Using [rug](https://docs.rs/rug/1.11.0/rug/index.html)


ayazhafiz:
Things to do:

- format floats without exponent suffix (https://gitlab.com/tspiteri/rug/-/issues/18)
- format floats as ints if we can convert them

ayazhafiz:
ptal @lukebhan

ayazhafiz:
Note: this would fail on "100/0" for example. We can fix it in a later pass.

PR #281, opened 2020-10-14
Install wasm-pack from cargo

ayazhafiz:
- Reduce failures due to GH API throttling
- Hopefully increase cache hits


ayazhafiz:
Looks like the latex image API is 500'ing

PR #283, opened 2020-10-15
Add diagnostic and hover language server

ayazhafiz:
This is a combination of 5 commits.

- Basic server implementation
- Implement diagnostic publishing
- Add support for hover
- Add tests for diagnostics
- Add tests for hover

### DIFF
--- a/editor/language_server/src/lsp.rs
+++ b/editor/language_server/src/lsp.rs
@@ -1,0 +1,207 @@
+//! Crate `slide_ls` implements a language server for [slide](libslide).
+
+#![deny(warnings)]
+#![deny(missing_docs)]
+#![doc(html_logo_url = "https://raw.githubusercontent.com/yslide/slide/master/assets/logo.png")]
+
+use libslide::*;
+
+use parking_lot::{MappedMutexGuard, Mutex, MutexGuard};
+use tower_lsp::jsonrpc::Result;
+use tower_lsp::lsp_types::*;
+use tower_lsp::{Client, LanguageServer, LspService, Server};
+
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::ops::Deref;
+
+mod services;
+mod shims;
+use shims::convert_diagnostics;
+
+#[cfg(test)]
+mod tests;
+
+// TODO(https://github.com/rust-lang/rust/issues/78003): used by pseudo-public providers.
+#[allow(private_in_public)]
+struct ProgramInfo {
+    source: String,
+    original: StmtList,
+    #[allow(unused)]
+    simplified: StmtList,
+}
+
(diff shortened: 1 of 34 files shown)


ayazhafiz:
@lukebhan - can you take a look. Do you want me to add tests?

editor/language_server/src/diagnostics.rs
```diff
@@ -0,0 +1,77 @@
...
+        .collect()
+}
+
+fn to_severity(diagnostic_kind: &s::DiagnosticKind) -> DiagnosticSeverity {
+    match diagnostic_kind {
+        s::DiagnosticKind::Error => DiagnosticSeverity::Error,
+        s::DiagnosticKind::Warning => DiagnosticSeverity::Warning,
+        s::DiagnosticKind::Note => DiagnosticSeverity::Information,
+        s::DiagnosticKind::Help => DiagnosticSeverity::Hint,
+    }
```
lukebhan, line 56: this seems useless. why dont you just replace severity with kind in a duagnostic above. also to/from would be better if not

editor/language_server/src/lsp.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    async fn initialized(&self, _params: InitializedParams) {
+        self.client
+            .log_message(MessageType::Info, "Slide language server initialized.")
+            .await;
+    }
+
+    async fn shutdown(&self) -> Result<()> {
+        Ok(())
```
lukebhan, line 107: does nothing?

editor/language_server/src/lsp.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    async fn did_open(&self, params: DidOpenTextDocumentParams) {
+        let TextDocumentItem { uri, text, version, .. } = params.text_document;
+        self.change(uri, text, Some(version)).await;
+    }
```
lukebhan, line 113: shouldnt this return something just or i guess its async so maybe not.

lukebhan (approved):

lukebhan:
> @lukebhan - can you take a look. Do you want me to add tests?

you probably should but i trust it work. i added some thins i noticed.

editor/language_server/src/diagnostics.rs
ayazhafiz: I can't, I have to change the slide diagnostic type to work with the other diagnostic type, which expects this other enum

editor/language_server/src/lsp.rs
ayazhafiz: Yep, we just have to implement it

editor/language_server/src/lsp.rs
ayazhafiz: yeah nothing to ret

PR #287, opened 2020-10-19
Completions

ayazhafiz:


editor/language_server/src/tests/completions.rs
```diff
@@ -0,0 +1,60 @@
...
+    var_completions: r"
+        a := a + a¦b
+                  ~@[a,c]
+        c := 1 + 2
+    "
+    var_completions_in_incomplete: r"
+        a := a + ¦
+                 ~@[a,c,d]
+        c := 1 + 2
+    "
```
ayazhafiz: This is broken, but I'm not sure why..

ayazhafiz: oh, it is because ~ should not be directly below ¦

editor/language_server/src/lsp.rs
```diff
@@ -62,12 +62,18 @@ impl SlideLS {
                 ..TextDocumentSyncOptions::default()
             },
         ));
+        let completion_provider = Some(CompletionOptions{
+            resolve_provider: Some(false),
+            trigger_characters: Some(vec!["=", "(", "[", "+", "-", "*", "/", "%", "^"].into_iter().map(String::from).collect()),
```
lukebhan, line 67: is there no better way to do this?

editor/language_server/src/lsp.rs
```diff
@@ -195,6 +201,20 @@ impl LanguageServer for SlideLS {
         self.close(&uri);
     }
 
+    async fn completion(&self, params: CompletionParams) -> Result<Option<CompletionResponse>> {
```
lukebhan, line 204: result<option> just looks funky, there's got to be something going on thats unneccessary.

editor/language_server/src/lsp.rs
```diff
@@ -195,6 +201,20 @@ impl LanguageServer for SlideLS {
...
+            position,
+        } = params.text_document_position;
+        self.client
+            .log_message(MessageType::Info, format!("Got completion request"))
+            .await;
+        let program_info = self.get_program_info(&uri);
+
+        let completions = services::get_completions(position, program_info.deref());
+        Ok(completions)
+    }
```
lukebhan, line 213: also you return result, but theres no case where it isn't ok.

editor/language_server/src/tests/mocks.rs
```diff
@@ -100,6 +100,19 @@ impl MockService {
...
+        &mut self,
+        uri: &Url,
+        position: Position,
+    ) -> Option<CompletionResponse> {
+        self.assert_ready();
+        let hover_resp = self
+            .send(text_document::definition::request(uri, position))
+            .await
+            .unwrap();
+        serde_json::from_value(hover_resp.get("result").unwrap().clone()).ok()
```
lukebhan: so if you can serialized to json, then this is valid. I'm not sur eI understand what this is doing.

editor/language_server/src/tests/completions.rs
lukebhan: brualski

lukebhan (approved):

editor/language_server/src/lsp.rs
ayazhafiz: nope

editor/language_server/src/lsp.rs
ayazhafiz: its required by the trait

editor/language_server/src/lsp.rs
ayazhafiz: its required by the trait

editor/language_server/src/tests/mocks.rs
ayazhafiz: hover_resp is json, we get the result key, then ask to convert that from json to a custom type

ayazhafiz:
don't have time to continue working on this rn, TODOs need to be fixed before landing but there are bigger priorities. Help is welcome!

editor/language_server/src/tests/completions.rs
```diff
@@ -0,0 +1,73 @@
...
+    //                 ~@[a,c]
+    //     c := 1 + 2
+    // "
+    // TODO: this should work
+    // var_completions_in_incomplete: r"
+    //     c := 1 + 2
+    //     a := a + ¦
+    //               ~@[c,a,d]
+    //     d := 1 + 2
+    // "
```
ayazhafiz, line 68: Must be fixed. This is basically just a job of improving the parser.

PR #303, opened 2020-11-17
Support for code actions

ayazhafiz:


ayazhafiz:
Blocking - going to go into libslide and make our diagnostics a bit more robust first (providing auto fixes in the diagnostic), which should make the work here easier.

